PR #3, opened 2025-03-09
Hammer changes

LordIdra:
- Swap nesting of hammer item and schema
- Change lore to explain how to use the hammer (this is a must in my experience)
- Require a block of the same material to use hammer on
- Change hammer recipe to be more intuitive (I think)
- Only damage and set cooldown if a recipe was attempted (this makes it clear if you have a valid recipe or not)
- Constant cooldown
- A few small code changes

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
@@ -76,7 +76,7 @@ public final class PylonItems {
                     .name("Stone Hammer")
                     .lore(
                             "Throw an item on top of a <yellow>stone block",
-                            "and <yellow>right click <gray>to use the hammer on it.",
+                            "and <yellow>right click</yellow> to use the hammer on it.",
                             "Higher tier hammers are more likely to succeed!"
                     )
                     .set(DataComponentTypes.ATTRIBUTE_MODIFIERS, ItemAttributeModifiers.itemAttributes()
@@ -109,7 +109,7 @@ public final class PylonItems {
                     .name("Iron Hammer")
                     .lore(
                             "Throw an item on top of an <yellow>iron block",
-                            "and <yellow>right click <gray>to use the hammer on it.",
+                            "and <yellow>right click</yellow> to use the hammer on it.",
                             "Higher tier hammers are more likely to succeed!"
                     )
                     .set(DataComponentTypes.ATTRIBUTE_MODIFIERS, ItemAttributeModifiers.itemAttributes()
@@ -142,7 +142,7 @@ public final class PylonItems {
                     .name("Diamond Hammer")
                     .lore(
                             "Throw an item on top of a <yellow>diamond block",
-                            "and <yellow>right click <gray>to use the hammer on it.",
+                            "and <yellow>right click</yellow> to use the hammer on it.",
                             "Higher tier hammers are more likely to succeed!"
                     )
                     .set(DataComponentTypes.ATTRIBUTE_MODIFIERS, ItemAttributeModifiers.itemAttributes()
--- a/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
@@ -23,7 +23,7 @@ import org.jetbrains.annotations.NotNull;
 import java.util.List;
 
 @SuppressWarnings("UnstableApiUsage")
-public class PylonItems {
+public final class PylonItems {
 
     private PylonItems() {
         throw new AssertionError("Utility class");
--- a/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
@@ -75,8 +75,9 @@ public final class PylonItems {
             new ItemStackBuilder(Material.STONE_PICKAXE)
                     .name("Stone Hammer")
                     .lore(
-                            "A hammer made of stone",
-                            "Useful as a weapon in a pinch"
+                            "Throw an item on top of a <yellow>stone block",
+                            "and <yellow>right click <gray>to use the hammer on it.",
+                            "Higher tier hammers are more likely to succeed!"
                     )
                     .set(DataComponentTypes.ATTRIBUTE_MODIFIERS, ItemAttributeModifiers.itemAttributes()
                             .addModifier(Attribute.ATTACK_SPEED, new AttributeModifier(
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
```diff
@@ -27,146 +28,147 @@
...
-            MiningLevel level,
-            float chance
-    ) implements Keyed {
-        @Override
-        public NamespacedKey getKey() {
-            return key;
+        if (anyRecipeAttempted) {
+            p.setCooldown(hammer, Schema.COOLDOWN_TICKS);
+            hammer.damage(1, p);
+            clickedBlock.getLocation().getWorld().playSound(clickedBlock.getLocation(), Sound.BLOCK_ANVIL_USE, 0.5F, 0.5F);
```
Seggan, line 155: Can't this be moved into the loop above right before the `break`?

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
```diff
@@ -27,146 +28,147 @@
...
+            ShapedRecipe recipe = new ShapedRecipe(id, template);
+            recipe.shape(
+                    " I ",
+                    " SI",
+                    "S  "
+            );
+            recipe.setIngredient('I', baseItem);
+            recipe.setIngredient('S', Material.STICK);
+            recipe.setCategory(CraftingBookCategory.EQUIPMENT);
+            RecipeTypes.VANILLA_CRAFTING.addRecipe(recipe);
```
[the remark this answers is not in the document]
LordIdra: I had it initially but then tried it and it looks absolutely cursed lol so I'd rather leave it

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -65,14 +65,15 @@ private PylonItems() {
...
             pylonKey("stone_hammer"),
-            Hammer.Item.class,
+            Hammer.class,
             new ItemStackBuilder(Material.STONE_PICKAXE)
                     .name("Stone Hammer")
                     .lore(
-                            "A hammer made of stone",
-                            "Useful as a weapon in a pinch"
+                            "Throw an item on top of a <yellow>stone block",
+                            "and <yellow>right click <gray>to use the hammer on it.",
```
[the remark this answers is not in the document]
LordIdra: Wait why? The intention was to highlight the 'instruction' part of it. If we make that yellow it would also make sense to highlight eg 'Throw an item'. Or is there different reasoning I'm missing?

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
LordIdra: oh yeah lmao

Seggan (approved):

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
```diff
@@ -27,146 +28,147 @@
...
 
-            Block clickedBlock = event.getClickedBlock();
-            World world = clickedBlock.getWorld();
+    public Hammer(Schema schema, ItemStack itemStack) {
+        super(schema, itemStack);
+    }
 
-            hammer.damage(1, p);
+    @Override
+    public void onUsedToRightClickBlock(@NotNull PlayerInteractEvent event) {
```
JustAHuman-xD, line 99: Didn't see this method name in that past PR or at least didn't think of it, but thats a pretty excessive name is it not xd?

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
```diff
@@ -27,146 +28,147 @@
...
+        event.setUseInteractedBlock(Event.Result.DENY);
 
-            if (event.getBlockFace() != BlockFace.UP) return;
+        Player p = event.getPlayer();
```
JustAHuman-xD, line 102: Starting this conversation asap, one of my least favorite things about lots of old plugin code bases is this weird notion that variable names should be as short as possible, or short like this, can we normalize this being `player` etc. (We should also probably make a code formatting/style guide at some point)

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
JustAHuman-xD, line 155: do we wanna use paper api/adventure sounds (also to make sounds more easily configurable?) This may be something for a later system but i definitely think configurable sounds is a must

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
LordIdra: it's been changed to onUsedToClickBlock since. when we release and update pylon-core next we can change this. I think that's still a better name than onClicked or something because it's so explicit

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
LordIdra: yeah good point. I'll change it to p. formatting should definitely be discussed soon

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
LordIdra: I've never used adventure API sounds so would rather just leave it and change this when I do configs for now

Intybyte (changes requested):

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -65,14 +65,15 @@ private PylonItems() {
...
+    public static final Hammer.Schema STONE_HAMMER = new Hammer.Schema(
             pylonKey("stone_hammer"),
-            Hammer.Item.class,
+            Hammer.class,
             new ItemStackBuilder(Material.STONE_PICKAXE)
                     .name("Stone Hammer")
                     .lore(
-                            "A hammer made of stone",
-                            "Useful as a weapon in a pinch"
+                            "Throw an item on top of a <yellow>stone block",
```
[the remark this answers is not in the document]
Seggan: All tags are auto closed at the end of the string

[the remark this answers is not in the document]
Intybyte: cool didn't know that, closing these then

Intybyte (approved):

PR #9, opened 2025-03-16
Add portable crafting table

OhmV-IR:
Does what the title says. works the same as if you just put down a crafting table and right-clicked it.
recipe is 4x compressed wood, each compressed wood is 9x logs of any type, but they have to be the same type, i.e 9x oak = 1 compressed and 9x dark oak = 1 compressed and these can be crafted and stacked together but 4x oak and 5x dark oak does not work
therefore portable crafting table costs 4 * 9 = 36 wood logs or 144 planks


src/main/java/io/github/pylonmc/pylon/base/items/CompressedWood.java
```diff
@@ -0,0 +1,31 @@
...
+import org.bukkit.inventory.ShapedRecipe;
+import org.bukkit.inventory.recipe.CraftingBookCategory;
+
+import java.util.List;
+
+public class CompressedWood extends PylonItemSchema {
+    private static final List<Material> WOODS = List.of(Material.OAK_WOOD, Material.BIRCH_WOOD, Material.SPRUCE_WOOD, Material.JUNGLE_WOOD, Material.ACACIA_WOOD, Material.DARK_OAK_WOOD, Material.MANGROVE_WOOD, Material.CRIMSON_STEM);
+    public CompressedWood(NamespacedKey id, Class<? extends SimplePylonItem> itemClass, ItemStack template){
+        super(id, itemClass, template);
+        for(var i = 0; i < WOODS.size(); i++){
```
Seggan, line 19: Instead of registering a recipe for every wood type I would suggest making a tab (probably in some PylonTags class) of the wood, and then use RecipeChoice.Material with that tag. Would also migitate the need for this entire class

Seggan (changes requested):

src/main/java/io/github/pylonmc/pylon/base/items/CompressedWood.java
OhmV-IR: I used the inbuilt logs tag so I didn't need the class, but I can still create it as a piece of infrastructure if you want

OhmV-IR:
**!! new recipe should work, but is untested ig, so wait for that before merge ** for some reason you need to connect to the mc session servers to join a localhost server :c

Seggan:
> for some reason you need to connect to the mc session servers to join a localhost server :c

Set offline mode to true

src/main/java/io/github/pylonmc/pylon/base/items/CompressedWood.java
Seggan: Nah that's fine

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -19,17 +21,15 @@
...
 
 import java.util.List;
 
 @SuppressWarnings("UnstableApiUsage")
 public final class PylonItems {
 
-    private PylonItems() {
-        throw new AssertionError("Utility class");
-    }
-
```
Seggan, line 32: This is standard in static-only classes

Seggan (changes requested):
Don't forget to delete the old `CompressedWood` class

LordIdra (approved):

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
Seggan, line 1: You deleted the whitespace between fields again

Seggan (changes requested):

Seggan (approved):

PR #11, opened 2025-03-16
Add portable echest

OhmV-IR:
Does what the title says. 
Recipe is 8x compressed obsidian and one eye of ender, where each obsidian is 9x obsidian so 8*9 = 72 obsidian in total.
shape is same as vanilla, except for compressed obsidian on outside instead of regular obsidian. I think that this might be a bit too expensive though because you can already carry echests around and this only saves the convenience of having to use silk touch and place it down and pick it back up every time.

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
@@ -29,7 +29,7 @@ import java.util.List;
 
 @SuppressWarnings("UnstableApiUsage")
 public final class PylonItems {
-    
+
     private PylonItems() {
         throw new AssertionError("Utility class");
     }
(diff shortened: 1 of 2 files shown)


Seggan (approved):
Check discord for my opinions on the recipe

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -19,17 +21,15 @@
...
+import org.bukkit.inventory.ShapedRecipe;
+import org.bukkit.inventory.recipe.CraftingBookCategory;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.List;
 
 @SuppressWarnings("UnstableApiUsage")
 public final class PylonItems {
 
-    private PylonItems() {
```
LordIdra, line 29: ?

OhmV-IR: put it back in the wrong place

Seggan (approved):

LordIdra (approved):

PR #13, opened 2025-03-24
Lore changes

LordIdra:


src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
```diff
@@ -35,34 +41,70 @@
...
 
+        private final Material baseBlock;
+        private final MiningLevel miningLevel;
+
         public Schema(
                 NamespacedKey key,
-                Class<? extends PylonItem<? extends PylonItemSchema>> itemClass,
-                ItemStack template,
+                String name,
+                Material item,
```
OhmV-IR, line 61: I think the name could be more descriptive, maybe `primaryCraftingItem` or something along those lines?

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -85,113 +88,50 @@ private PylonItems() {
...
-                    )
-                    .build(),
-            MiningLevel.STONE,
+            "Stone Hammer",
+            Material.COBBLESTONE,
             Material.STONE,
-            new RecipeChoice.MaterialChoice(Tag.ITEMS_STONE_TOOL_MATERIALS)
+            Material.STONE_PICKAXE,
+            MiningLevel.STONE,
+            (1.0 / 3) - 4,
```
OhmV-IR, line 96: 1/3 - 4 is -11/3 but this should be a positive quantity I think

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -85,113 +88,50 @@ private PylonItems() {
...
-                    )
-                    .build(),
-            MiningLevel.DIAMOND,
+            "Diamond Hammer",
+            Material.DIAMOND,
             Material.DIAMOND_BLOCK,
-            new RecipeChoice.MaterialChoice(Material.DIAMOND)
+            Material.DIAMOND_PICKAXE,
+            MiningLevel.DIAMOND,
+            1 - 4,
```
OhmV-IR, line 120: same here, why negative?

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -211,11 +151,294 @@ private PylonItems() {
                     .build()
     );
 
+    public static final PylonItemSchema FERRODURALUM_ORE = new SimpleItemSchema<>(
```
OhmV-IR, line 154: I could copy the lore changes into my ferroduralum branch so that this merge wouldn't be waiting for a fix for the ferroduralum items. The armor is being a bit of a problem rn since when I set any attribute, it applies it as an entity attribute so it works when its in your hand and equipped as an armor piece instead of just when its equipped as an armor piece

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
OhmV-IR: I see that it was the same way in the old one, but I'd like to fix it here

OhmV-IR (changes requested):
Just chore fixes, but this PR will have to wait on the ferroduralum armor fix unless the changes are copied to that branch

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
Seggan: It can't really be fixed, we want to set a certain value which is why the `- 4` exists, to cancel out the default

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
Seggan: Same reason

Seggan (approved):

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
LordIdra: Nah dw it's fine we can wait on ferroduralum items. There isn't really a rush to merge this

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
LordIdra: Because this gets added to the cooldown rather than being an absolute value. It's kinda annoying. There isn't a way to directly set the cooldown you can only add/subtract

src/main/java/io/github/pylonmc/pylon/base/items/Hammer.java
LordIdra: Maybe but I think it's obvious enough to not be worth changing. If you're looking at a call site you already see that Material.DIAMOND is passed in, which immediately indicates what the item is. If you're looking at the method itself you can just check where it's used and see it's just used in the recipe

OhmV-IR (approved):
LGTM

PR #15, opened 2025-03-25
Add health talismans

OhmV-IR:
As opposed to the healing items, these are passive talismans that increase your max health while they are in your inventory. They are not stackable in the sense that you can't have 2 basic talismans in your inv for twice the effect, but you can have one simple and one advanced .etc and get the effects of each to compound together. I think this is fine because you make a tradeoff between  having the extra item slots and the additional health. 
**Recipes**
*simple talisman*
8x glistering melon slice(this is the ingredient used in health potions, therefore it makes sense to use it here)
1x redstone(idk, just felt like it, redstone is pretty underused in vanilla progression so giving it a purpose is nice)
+2 max hp
*advanced talisman*
9x simple talismans(or 72 glistering melon slices and 9 redstone)
+6 max hp
*ultimate talisman*
9x advanced talismans(or 648 glistering melon slices and 81 redstone)
+10 max hp
Depends on Core [#40](https://github.com/pylonmc/pylon-core/pull/40)


LordIdra:
I do think the scaling should be more like 4 -> 7 -> 10 maybe, just so it's more worth crafting the basic ones and not just rushing the advanced ones?

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
JustAHuman-xD, line 1: Indentation levels if possible should be reduced, also not a fan of having a ticker per pylon item, seems vastly inefficient. If you have a ton of items that tick and all of them with tickers, then you'll have each one checking all of the inventory for no reason. It should be one ticker that checks the whole inventory and calls some sort of interface (I thought that was discussed already?)

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,96 @@
...
+                    .attributeLine("Max health increase", 6, Quantity.HEARTS)
+                    .arrow().text("Passive effect while in inventory").newline()
+                    .arrow().text("Does not stack"))
+            .set(DataComponentTypes.MAX_STACK_SIZE, 1)
+            .build();
+    public static final ItemStack ULTIMATE_TALISMAN_STACK = new ItemStackBuilder(Material.BUDDING_AMETHYST)
+            .name("Ultimate Healing Talisman")
+            .lore(new LoreBuilder()
+                    .attributeLine("Max health increase", 10, Quantity.HEARTS)
+                    .arrow().text("Passive effect while in inventory").newline()
```
JustAHuman-xD, line 36: Is there a reason the itemstacks are in here vs PylonItems like the rest of them?
Also static fields should be above any class fields

JustAHuman-xD (changes requested):

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
OhmV-IR: from idra on old pr:
Ok - no. What I suggested is not using this interface for the health talismans. Instead, create a separate ticker that only runs every 40 ticks or whatever and does this:

Checks every player's inventory for a health talisman
If found, check for health-talisman-active key or whatever
If that key is not found or set to false, set it to true, and increase the player's health
If health talisman not found, check for health-talisman-active key or whatever
If that key is found and set to true, reduce the player's health by however many hearts and set it to false
THis is the only way I can see to guarantee this works regardless of interference from other plugins or whatever

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
OhmV-IR: For making the recipes inside of the constructor

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
LordIdra: That's not what justin is saying, what I've been getting at is **don't use a ticker per item, instead use a single ticker.**

LordIdra: So don't need an exit/enter event, instead just have a tick method and have **one single ticker overall** which handles **all** items

LordIdra: But lmk if anything is still unclear

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
LordIdra: You can still use them with eg. Items.SIMPLE_TALISMAN_STACK tho

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
LordIdra: Ok sorry for 4th comment - but I realised I'm wrong here and was conflating this with the inventory ticker suggestion I made. What you've got is fine I think. Justin refer to core#15 for this

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
```diff
@@ -0,0 +1,58 @@
...
+    public void run() {
+        for (Player player : PylonBase.getInstance().getServer().getOnlinePlayers()) {
+            boolean foundItem = false;
+            for (ItemStack itemStack : player.getInventory()) {
+                PylonItem<?> pylonItem = PylonItem.fromStack(itemStack);
+                if (!(pylonItem instanceof HealthTalisman.HealthTalismanItem)) {
+                    continue;
+                }
+                HealthTalisman.HealthTalismanItem talisman = ((HealthTalisman.HealthTalismanItem) pylonItem);
+                if (!player.getPersistentDataContainer().has(healthBoostedKey)) {
```
LordIdra, line 28: These if statements all use the same values (like player.getPersistentDataContainer().has(healthBoostedKey)) - you should store the values in a local variable to make the code more readable

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class HealthTalisman extends PylonItemSchema {
+    private static boolean recipesRegistered = false;
+    public int healthAmount;
+
+    public HealthTalisman(NamespacedKey id, Class<? extends PylonItem<? extends HealthTalisman>> itemClass
+            , ItemStack template, int healthAmount, Function<ItemStack, ShapedRecipe> recipeFunc) {
+        super(id, itemClass, template);
+        this.healthAmount = healthAmount;
+        assert template.getMaxStackSize() == 1;
```
LordIdra, line 24: I would use an exception here instead of an assert, and supply a clear error message

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,44 @@
...
+import org.bukkit.inventory.ShapedRecipe;
+import org.bukkit.inventory.recipe.CraftingBookCategory;
+
+import java.util.function.Function;
+
+public class HealthTalisman extends PylonItemSchema {
+    private static boolean recipesRegistered = false;
+    public int healthAmount;
+
+    public HealthTalisman(NamespacedKey id, Class<? extends PylonItem<? extends HealthTalisman>> itemClass
```
LordIdra, line 20: Just some formatting convention stuff, when doing multi-line parameter lists, have each parameter on its own line

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,44 @@
...
+import org.bukkit.Material;
+import org.bukkit.NamespacedKey;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.ShapedRecipe;
+import org.bukkit.inventory.recipe.CraftingBookCategory;
+
+import java.util.function.Function;
+
+public class HealthTalisman extends PylonItemSchema {
+    private static boolean recipesRegistered = false;
```
LordIdra, line 17: Unused

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,44 @@
...
+            super(schema, itemStack);
+            this.schema = schema;
+        }
+
+        public int getHealthIncrease(){
+            return schema.healthAmount;
+        }
+    }
+
+    public static interface HealthTalismanItem {
```
LordIdra, line 41: Doesn't need static

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,44 @@
...
+        public int getHealthIncrease();
```
LordIdra, line 42: Doesn't need public

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static class Item extends PylonItem<HealthTalisman> implements HealthTalismanItem {
+        HealthTalisman schema;
+
+        public Item(HealthTalisman schema, ItemStack itemStack) {
+            super(schema, itemStack);
+            this.schema = schema;
+        }
+
+        public int getHealthIncrease(){
```
LordIdra, line 36: I would omit this and just do `item.schema.healthAmount`, rather than having a whole interface for it.

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
```diff
@@ -0,0 +1,58 @@
...
+                    player.getAttribute(Attribute.MAX_HEALTH).addModifier(new AttributeModifier(
+                            healthBoostedKey,
+                            talisman.getHealthIncrease(),
+                            AttributeModifier.Operation.ADD_NUMBER
+                    ));
+                    player.getPersistentDataContainer().set(healthBoostedKey, PersistentDataType.INTEGER, talisman.getHealthIncrease());
+                    foundItem = true;
```
LordIdra, line 35: You can break when the item is found (in all branches)

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
```diff
@@ -0,0 +1,58 @@
...
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.persistence.PersistentDataType;
+import org.bukkit.scheduler.BukkitRunnable;
+
+public class HealthTalismanTicker extends BukkitRunnable {
+    private static final NamespacedKey healthBoostedKey = new NamespacedKey(PylonBase.getInstance(), "health_boosted");
+
+    @Override
+    // Suppresses warnings from doing PDC.has() and then .get() and assuming .get is not null, and assuming that the player has the max health attribute
+    @SuppressWarnings("DataFlowIssue")
```
LordIdra, line 18: Instead of this, I would store player.getPersistentDataContainer().get(healthBoostedKey, PersistentDataType.INTEGER), and then have an explicit null check where you continue if it's null. This would be a lot clearer and avoid the suppression here

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
```diff
@@ -0,0 +1,58 @@
...
+import org.bukkit.NamespacedKey;
+import org.bukkit.attribute.Attribute;
+import org.bukkit.attribute.AttributeModifier;
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.persistence.PersistentDataType;
+import org.bukkit.scheduler.BukkitRunnable;
+
+public class HealthTalismanTicker extends BukkitRunnable {
+    private static final NamespacedKey healthBoostedKey = new NamespacedKey(PylonBase.getInstance(), "health_boosted");
```
LordIdra, line 14: Maybe `talisman_health_boosted` to avoid clashes with other items?

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -690,6 +683,82 @@ private PylonItems() {
...
     );
+
+    public static final HealthTalisman SIMPLE_HEALTH_TALISMAN = new HealthTalisman(
+            pylonKey("simple_healing_talisman"),
+            HealthTalisman.Item.class,
+            new ItemStackBuilder(Material.AMETHYST_SHARD)
+                    .name("Simple Healing Talisman")
+                    .lore(new LoreBuilder()
+                            .attributeLine("Max health increase", 6, Quantity.HEARTS)
+                            .arrow().text("Passive effect while in inventory").newline()
```
LordIdra, line 694: Should have a space before 'Passive'. Same with 'Does' on the next line. (And obviously goes for all the talismans)

src/main/java/io/github/pylonmc/pylon/base/PylonItems.java
```diff
@@ -690,6 +683,82 @@ private PylonItems() {
...
+            }
+    );
+
+    public static final HealthTalisman ADVANCED_HEALTH_TALISMAN = new HealthTalisman(
+            pylonKey("advanced_healing_talisman"),
+            HealthTalisman.Item.class,
+            new ItemStackBuilder(Material.AMETHYST_CLUSTER)
+                    .name("Advanced Healing Talisman")
+                    .lore(new LoreBuilder()
+                            .attributeLine("Max health increase", 6, Quantity.HEARTS)
```
LordIdra, line 719: 10, not 6

LordIdra (changes requested):
Looks like the logic is working well which is really good. Just some small comments here and there.

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
OhmV-IR: doing a null check just to continue anyways seems very pointless and more confusing if anything. Also if statements are not free, so I'd prefer to just suppress the warnings. The comment provides the clarity on why the warnings are suppressed

LordIdra: If statements are not free, but compiler optimisations are very much free and it's probably not worth thinking about performance in that regard. The PDC checks are probably like 100x slower than the ifs. Anyway what I mean is that you can store playerPDC.get(healthBoostedKey, PersistentDataType.INTEGER) in a variable, which stops you from having to retrieve it multiple times. Then, just adding a null check for that is more direct because you're explicitly telling the compiler that you're checking if the variable is null - and this allows you to skip  the playerPDC.has() check as well. In fact you can just replace `if (!playerPDC.has(healthBoostedKey)) {` with `if (healthBoost == null) {`

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
```diff
@@ -0,0 +1,60 @@
...
+import org.bukkit.NamespacedKey;
+import org.bukkit.attribute.Attribute;
+import org.bukkit.attribute.AttributeModifier;
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.persistence.PersistentDataContainer;
+import org.bukkit.persistence.PersistentDataType;
+import org.bukkit.scheduler.BukkitRunnable;
+
+public class HealthTalismanTicker extends BukkitRunnable {
```
LordIdra, line 14: Sorry just remembered I forgot to comment this - I think this should be a subclass of HealthTalisman to keep with the convention of placing all the stuff related to an item in one class

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalismanTicker.java
Seggan: yeah null branches are like 2 machine code ops

LordIdra:
Just 3 more things, I think you might have missed them because github collapsed them

OhmV-IR:
> Just 3 more things, I think you might have missed them because github collapsed them

Which three things? I can't find them

LordIdra (approved):

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,89 @@
...
+                    PylonItem<?> pylonItem = PylonItem.fromStack(itemStack);
+                    if (!(pylonItem instanceof HealthTalisman.Item talisman)) {
+                        continue;
+                    }
+                    if (playerHealthBoost == null) {
+                        player.getAttribute(Attribute.MAX_HEALTH).addModifier(new AttributeModifier(
+                                healthBoostedKey,
+                                talisman.schema.healthAmount,
+                                AttributeModifier.Operation.ADD_NUMBER
+                        ));
```
Intybyte, line 65: Instead of creating the attribute modifier each time i think it should be created inside the HealthTalisman craft with a getter access

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,89 @@
...
+                        playerPDC.set(healthBoostedKey, PersistentDataType.INTEGER, talisman.schema.healthAmount);
+                        foundItem = true;
+                    }
+                    else if (playerHealthBoost < talisman.schema.healthAmount) {
```
Intybyte, line 69: identation

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,89 @@
...
+                boolean foundItem = false;
+                PersistentDataContainer playerPDC = player.getPersistentDataContainer();
+                Integer playerHealthBoost = playerPDC.get(healthBoostedKey, PersistentDataType.INTEGER);
+                for (ItemStack itemStack : player.getInventory()) {
+                    PylonItem<?> pylonItem = PylonItem.fromStack(itemStack);
+                    if (!(pylonItem instanceof HealthTalisman.Item talisman)) {
+                        continue;
+                    }
+                    if (playerHealthBoost == null) {
+                        player.getAttribute(Attribute.MAX_HEALTH).addModifier(new AttributeModifier(
```
Intybyte, line 61: can we save `player#getAttribute(Attribute.MAX_HEALTH)` to a variable outside the for cycle?

Intybyte (changes requested):

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
OhmV-IR: not sure what you were referring to, I reran autoformat on the file so that might've fixed whatever it was

OhmV-IR:
hold off on merge, there's a bug somewhere that is causing this not to work, I'll investigate more later

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
Intybyte: meant `} else if ...`

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.function.Function;
+
+public class HealthTalisman extends PylonItemSchema {
+    public int healthAmount;
+    private static final NamespacedKey healthBoostedKey = new NamespacedKey(PylonBase.getInstance(), "talisman_health_boosted");
+    private AttributeModifier healthModifier = new AttributeModifier(
+            healthBoostedKey,
+            healthAmount,
+            AttributeModifier.Operation.ADD_NUMBER
+    );
```
Intybyte, line 28: No need to allow change to healthAmount, you can make it final, same for healthModifier, you can also technically nuke the getter in favor of public final in healthModifier too

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -19,29 +19,23 @@
 import java.util.function.Function;
 
 public class HealthTalisman extends PylonItemSchema {
-    public int healthAmount;
+    public final int healthAmount = getSettings().getOrThrow("max-health-boost", Integer.class);
```
Intybyte, line 22: Won't this make all talisman provide the same health amount?

Intybyte (approved):

src/main/java/io/github/pylonmc/pylon/base/items/HealthTalisman.java
```diff
@@ -19,29 +19,23 @@
...
             healthAmount,
             AttributeModifier.Operation.ADD_NUMBER
     );
 
     public HealthTalisman(NamespacedKey id,
                           Class<? extends PylonItem<? extends HealthTalisman>> itemClass,
-                          ItemStack template,
-                          int healthAmount,
+                          Function<NamespacedKey, ItemStack> template,
                           Function<ItemStack, ShapedRecipe> recipeFunc) {
```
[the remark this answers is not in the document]
LordIdra: What's the advantage to doing this?

LordIdra (approved):
Happy to merge when formatting fixed

PR #23, opened 2025-04-09
Change altar/pedestal to use new entity holder block

LordIdra:
Companion to core#65


Seggan (approved):

LordIdra:
Simply merging this into magic altar so it's 1 PR

PR #32, opened 2025-04-17
Add block names

Seggan:
Required for WAILA

### DIFF
--- a/src/main/resources/lang/pylonbase/en.yml
+++ b/src/main/resources/lang/pylonbase/en.yml
@@ -156,8 +156,8 @@ item:
   pedestal:
     name: "Pedestal"
     lore: |-
-      <arrow> <isn>Right click</isn> to set the displayed item
-      <arrow> <isn>Shift right click</isn> to rotate the item
+      <arrow> <insn>Right click</insn> to set the displayed item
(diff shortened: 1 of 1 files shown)


src/main/resources/lang/pylonbase/en.yml
```diff
@@ -211,6 +211,14 @@ item:
       <arrow> Passive effect while in inventory
       <arrow> Does not stack
 
+block:
+  grindstone: "Grindstone"
+  grindstone_handle: "Grindstone Handle"
+  magic_altar: "Magic Altar"
+  magic_pedestal: "Magic Pedestal"
+  pedestal: "Pedestal"
+  sprinkler: "Sprinkler"
```
LordIdra, line 220: Missing phantom block. Also just to confirm, you can use placeholders here right?

LordIdra (changes requested):

src/main/resources/lang/pylonbase/en.yml
LordIdra: (Actually phantom goes in core sorry - but still)

Seggan: Yeah, you can use placeholders

LordIdra (approved):

PR #34, opened 2025-04-18
Recipe and item refactor

LordIdra:
## Summary
(well, 'summary')
- Items, blocks, and entities are now registered in static blocks next to each item/block/entity definition
- Item recipes are now all added in the static blocks after registration
- Changed config names for beheading sword to be more clear
- Lots of formatting fixes
- Standardised shaped recipe formatting so all three layers of the shape are on one line
- Renamed various `Item` classes to have a more specific name (eg `BeheadingSwordItem`)
- Items that modify the item stack now do `template.xyz` after calling super, to avoid the weird situation we had in hammer with items, and moreover make a standard way to do this
- Hammer now has a getRecipe() function. This is my compromise between overriding onRegister to register the recipe (not nice) and duplicating the recipe creation code for each hammer item (really not nice)
- Renamed healing talismans to health talismans (we were using both names in different places)
- Config field values now match their config name (eg `max-health-boost` config value has a corresponding field called maxHealthBoost)
- Functions/constructors that override or call super now have matching parameters to their super name (eg, template renamed to templateSupplier in several items)
- Removed redundant schemas that could just be PylonItemSchema in all the portable items
- Custom schemas by and large now supply their item/block class rather than taking it as a parameter (for example, see watering can)
- Added constants for default cooking times
- All items now use defaultBuilder instead of some of them using translatableName and some of them using defaultBuilder

## Language issue
Currently using `lore: []` just prints out the translation key, instead of not adding any lore. This PR won't display the correct lore until this is fixed

## Future work
(will do in next few PRs)
- SimpleBlockPlacer item will help reduce boilerplate
- We are using different schema/item nesting order in different items. This should be standardised to Item { ... Schema ... } as agreed at some point on discord
- Many recipes need adding or modifying. I'll do this after re-merging the mixing pot and magic intermediates PRs


Seggan (approved):
Might wanna rename the talismans to something like ex `health_talisman_advanced` for parity with hammers and better alphabetization

Seggan:
You forgot to update the translation keys

Issue #42, opened 2025-04-22
Some way to convert bamboo to string


LordIdra:
maybe a grindstone recipe?

PR #49, opened 2025-05-12
bamboo to string recipe (ref: #42)

glomdom:
references issue #42

added a recipe for the grindstone which takes in 3 bamboo and outputs 1 string, taking 3 cycles

im also unsure if this is a good way to structure where to add this recipe, as i couldn't find any already existing recipes that use a pylon block and output a vanilla item, so i added it where the block is registered. imo this isn't very good, but i couldn't decide on what to do so ill leave that to one of the main contributors

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/PylonBlocks.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/PylonBlocks.java
@@ -66,7 +66,7 @@ public final class PylonBlocks {
         GRINDSTONE.register();
         Grindstone.Recipe.RECIPE_TYPE.addRecipe(new Grindstone.Recipe(
             pylonKey("string_from_bamboo"),
-            new RecipeChoice.ExactChoice(new ItemStack(Material.BAMBOO)),
+            new RecipeChoice.MaterialChoice(Material.BAMBOO),
             3,
             new ItemStack(Material.STRING),
             3,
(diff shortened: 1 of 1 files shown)


Seggan (approved):
Untested

src/main/java/io/github/pylonmc/pylon/base/PylonBlocks.java
```diff
@@ -61,6 +64,17 @@ private PylonBlocks() {
     );
     static {
         GRINDSTONE.register();
+        Grindstone.Recipe.RECIPE_TYPE.addRecipe(new Grindstone.Recipe(
+            pylonKey("string_from_bamboo"),
+            new RecipeChoice.ExactChoice(new ItemStack(Material.BAMBOO)),
```
OhmV-IR, line 69:
```suggestion
            new RecipeChoice.MaterialChoice(Material.BAMBOO),
```
cleaner and does the same thing

src/main/java/io/github/pylonmc/pylon/base/PylonBlocks.java
```diff
@@ -61,6 +64,17 @@ private PylonBlocks() {
...
+            3,
```
OhmV-IR, line 70:
```suggestion
            4,
```
OPTIONAL: 3 is an awkward number in minecraft, I prefer to just work in powers of 2 where possible / makes sense, but this is ultimately my preference so it's your choice whether you take this

OhmV-IR (changes requested):

OhmV-IR (approved):
LGTM, tested in-game

PR #51, opened 2025-05-12
Organize

Seggan:
Organized the item classes a bit


LordIdra:
Not gonna block but I think this should wait until we have the guide and categories so we can organise the items by category

Seggan:
At least for now? Guide's been taking a bit

LordIdra (approved):
k

PR #53, opened 2025-05-13
Bigger barrel

Seggan:
Not *quite* bottomless. Created as a test for inventory saving


src/main/java/io/github/pylonmc/pylon/base/items/BottomlessBarrel.java
```diff
@@ -0,0 +1,91 @@
...
+    public static final class BottomlessBarrelItem extends PylonItem<PylonItemSchema> implements BlockPlacer {
+
+        public BottomlessBarrelItem(@NotNull PylonItemSchema schema, @NotNull ItemStack stack) {
+            super(schema, stack);
+        }
+
+        @Override
+        public @NotNull PylonBlockSchema getBlockSchema() {
+            return PylonBlocks.BOTTOMLESS_BARREL;
+        }
```
OhmV-IR, line 83: Think about providing this in the schema constructor so that this class could be reused to create different tiers of barrels

src/main/java/io/github/pylonmc/pylon/base/items/BottomlessBarrel.java
```diff
@@ -0,0 +1,91 @@
...
+
+        @Override
+        public @NotNull Map<@NotNull String, @NotNull Component> getPlaceholders() {
+            return Map.of("size", Component.text(PylonBlocks.BOTTOMLESS_BARREL.size));
```
OhmV-IR, line 87:
```suggestion
            return Map.of("size", Component.text(getBlockSchema().size));
```
less hardcode

src/main/java/io/github/pylonmc/pylon/base/util/gui/DefaultItems.java
```diff
@@ -0,0 +1,32 @@
+package io.github.pylonmc.pylon.base.util.gui;
+
+import io.github.pylonmc.pylon.core.item.builder.ItemStackBuilder;
+import org.bukkit.Material;
+import xyz.xenondevs.invui.item.Item;
+import xyz.xenondevs.invui.item.impl.SimpleItem;
+
+public final class DefaultItems {
```
OhmV-IR, line 8: Instead of DefaultItems which sounds a bit too global, maybe just GUIItems to give a bit more of an indicator to the purpose

OhmV-IR (changes requested):
Bug 1: Putting items in and then breaking the barrel and replacing it causes those items to be lost
Either drop the items inside when the barrel is broken or carry them in the barrel and then allow them to be reaccessed after it is placed down again
Got this to trigger just with vanilla blocks in there

Seggan:
Hehe oops, you are the best Ohm

OhmV-IR (approved):
LGTM CW&IG

LordIdra (changes requested):
Just thinking about the UX, maybe it would be better to have 'next page' and 'previous page' here? I'd go insane if I had to scroll a bunch to find the item I'm looking for. Also maybe call it something different, since from the name I thought it would be an endless barrel? Otherwise looks good

LordIdra (approved):
Sorry missed this when reviewing stuff earlier. Looks really good. One thing I'd suggest is to add a page indicator like `Next page (1/4)` on the next/previous page items like SF does it. Even if not for this, probably useful in other UIs. But not blocking on it.

PR #55, opened 2025-05-17
Add structure locators

OhmV-IR:
Compass, right click to point towards the nearest structure of that type.
Code is 100% reusable and should work for every structure in the game, but I only did nether fortresses and desert pyramids, just for testing mainly.


src/main/java/io/github/pylonmc/pylon/base/items/StructureLocator.java
```diff
@@ -0,0 +1,81 @@
...
+    private String keyToDisplayStr(String messyKey) {
+        StringBuilder strBuilder = new StringBuilder();
+        for (char character : messyKey.toCharArray()) {
+            if (character != '_') {
+                strBuilder.append(character);
+            } else {
+                strBuilder.append(' ');
+            }
+        }
+        return strBuilder.toString();
```
Seggan, line 67: `messyKey.replace('_', ' ')`

src/main/resources/lang/pylonbase/en.yml
```diff
@@ -259,6 +259,20 @@ item:
       <arrow> Makes crops, saplings, sugar cane, and cactus go brrr
       <arrow> <attr:Range:blocks>%range%</attr>
 
+  structure_locator_desert_pyramid:
+    name: "Structure Locator"
+    lore: |-
+      <arrow> Navigates to the nearest %struct%
+      <arrow> <insn>Right click</insn> to refresh
+      <arrow> <attr:Range:chunks>%range%</attr>
```
Seggan, line 267: `chunks` is not a registered `Quantity`, either add the quantity to core or use the `blocks` quantity

src/main/resources/lang/pylonbase/en.yml
```diff
@@ -339,3 +353,5 @@ message:
       1: "you find that a seemingly living gelatinous substance has formed in the pack"
       2: "a strange, glowing liquid spills out"
       3: "full of anticipation, you are disappointed to find that nothing happens"
+  structurecompass_find_fail: "<red>Failed to find a %struct% within %range% chunks"
+  structurecompass_refreshed: "<green>Compass reprogrammed!"
```
Seggan, line 357: Can be put under one key ex `structure_compass.find_fail`

Seggan (changes requested):

LordIdra:
This makes the server lag like CRAZY when it's right clicked. I think because it's generating new chunks. Also - I think this would be better to have as a one-time item, because otherwise this is potentially very overpowered (we run out of end cities within a few weeks on MM as is even with a large world - this would make that a few days)

OhmV-IR:
> Still very slow. Is there a way to do this without generating new chunks, ie by predicting where stuff will spawn? If not I don't think we should add this item - you could also pass false to the structure locate function to not generate new chunks but this would make it much less useful (and potentially confusing)

To predict where structures would be we'd basically have to do the generating ourselves which isn't all that easy and it would easily break with new versions of mc. There just no easy way over the balancing and performance hurdles, so I'm going to close this as not planned.

PR #63, opened 2025-05-24
Yeet block schemas

LordIdra:


Seggan (approved):

OhmV-IR (changes requested):
LGTM CW but not IG: Dimensional barrel bug: breaking it doesn't drop the items inside of it and the items inside of it do not persist through server restarts.
Mixing pot also seems not to work, just doesn't craft the item but I might just have a skill issue
Didn't review grindstone, magic altar, or mixing pot classes since the diffs were too cooked and I just tested them in-game instead.

PR #73, opened 2025-06-02
Merge waila entries and add fluid names

LordIdra:
- Change language file in line with https://github.com/pylonmc/pylon-core/pull/131
- Add fluid names to language file and use in WAILA + placeholders


OhmV-IR (approved):
LGTM CW & IG, but note that the fluid WAILAs don't refresh in real-time, but that is a thing for another time

Seggan (approved):

PR #86, opened 2025-06-26
Add food processor

OhmV-IR:
Adds food processor multiblock, some nutritious foods incoming as well, note that I tagged this version of it as the "simple" food processor as I think it would make for a good energy machine later down the line which would be the "advanced+" version of the food processor. Also fixes bug in dimensional barrel where items would disappear upon server restart


Seggan (changes requested):

src/main/java/io/github/pylonmc/pylon/base/items/multiblocks/FoodProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+        public static final RecipeType<FoodProcessor.SimpleRecipe> RECIPE_TYPE = new RecipeType<>(
+                new NamespacedKey(PylonBase.getInstance(), "simple_food_processor")
+        );
+
+        static {
+            PylonRegistry.RECIPE_TYPES.register(RECIPE_TYPE);
+        }
+    }
+
+    public static class FoodProcessorHandle extends PylonBlock implements PylonInteractableBlock {
```
LordIdra, line 119: I would suggest making this a separate class, not nested, for consistency with how this is handled elsewhere( grindstone / grindstone handle for example)

LordIdra (changes requested):
Looks fine to me codewise but I'd suggest 2 changes
1) 'salad' should definitely not be a part of base IMO for the simple reason that food is probably best left to an addon (think exotic garden / gastronomicon). I think this is a fine thing to add if we lean more into the fantasy food side of things like the golden pie
2) Similar vein but 'food processor' feels like a slightly weird/generic name. If we focus more on fantasy foods it might make sense to call this something else? Not quite sure what though

LordIdra:
Ready for review now?

src/main/java/io/github/pylonmc/pylon/base/BaseItems.java
```diff
@@ -75,6 +56,7 @@ private BaseItems() {
     //<editor-fold desc="Dusts" defaultstate=collapsed>
     public static final ItemStack COPPER_DUST = ItemStackBuilder.pylonItem(Material.GLOWSTONE_DUST, BaseKeys.COPPER_DUST)
             .build();
+
```
LordIdra, line 59: Can we undo this whitespace stuff?

LordIdra (changes requested):
The golden pie seems a little ridiculous :sweat_smile: wouldn't it be better to split it into several different foods, each one granting a different set of bonuses? Also placeholders are broken I think
image

LordIdra:
Also, mystical food enhancer does not seem to work?
image

OhmV-IR:
this pr needs a lot of work to get updated(doesn`t even compile with all the refactors that have happened since :skull:). will try to do at somepoint between all the reviews but I`ve marked it as a draft for now

OhmV-IR:
Going to close this for the following reasons:
1. an addon can do this much better
2. there is really no good place in the current progression for this
3. there's no way I have time to get this back up to speed anytime soon

PR #99, opened 2025-07-12
Migrate fluids to core

LordIdra:
See https://github.com/pylonmc/pylon-core/pull/169/files


OhmV-IR (changes requested):
this fails to compile. seems that it can't find PylonGuide and FluidConnectionPoint

LordIdra:
oh make sure you check out the core version

Seggan (approved):

LordIdra:
> > Member
> 
> did that and gradlew cleaned too

Compiling fine for me with latest. Checked and double checked I'm synced with both branches

PR #101, opened 2025-07-13
 add wither sword

OhmV-IR:
shoots wither skulls out of sword on right click, has super variant

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/BaseItems.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/BaseItems.java
@@ -1681,6 +1681,9 @@ public final class BaseItems {
     public static final ItemStack REACTIVATED_WITHER_SKULL = ItemStackBuilder.pylonItem(Material.WITHER_SKELETON_SKULL, BaseKeys.REACTIVATED_WITHER_SKULL)
             .set(DataComponentTypes.MAX_DAMAGE, Settings.get(BaseKeys.REACTIVATED_WITHER_SKULL).getOrThrow("durability", ConfigAdapter.INT))
             .set(DataComponentTypes.DAMAGE, 0)
+            .set(DataComponentTypes.USE_COOLDOWN, UseCooldown.useCooldown(0f)
+                    .cooldownGroup(BaseKeys.REACTIVATED_WITHER_SKULL)
+                    .build())
             .build();
     static {
         PylonItem.register(ReactivatedWitherSkull.class, REACTIVATED_WITHER_SKULL);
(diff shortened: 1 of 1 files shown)


LordIdra (approved):
Haven't tested ingame but looks good

LordIdra (changes requested):
- _Very_ overpowered as is because it can break a huge amount of blocks for free. I'd suggest adding a cooldown to it, and also taking durability whenever a skull is fired
- Currently seems to shoot the skulls from your feet, so you have to be looking up to shoot a skull otherwise it just explodes immediately
- More a personal preference thing, but I don't think 'sword' suits this item very well. Something like an actual wither skull item ('Reactivated Wither Skeleton Skull') or something similar seems like it might be better - this could be quite cool if paired with the idea above of reducing durability when a skull is fired

Seggan:
Why "not work" I don't see you setting a cooldown anywhere

LordIdra (changes requested):
A few nitpicks
- Skull name is aqua, you can add `<white>` in the item name to override this
- 'Hyperactive' with skull doesn't really fit here. Maybe Hyper_activated_? Also the files and keys and such are still 'super' not 'hyperactive'
- I would suggest nerfing the cost a bit, maybe 1 nether star instead of 3?
- Hyper wither skull is IMO less useful than the reactivated one because the explosions are the same size but the head slows down quickly. Not sure how this can be fixed though

OhmV-IR:
The only difference between the hyper and the regular one is that the hyper one has withering effect enabled for the entities hit by explosion, not sure why the skull is seemingly travelling slower, but I'll make the cooldown on the two the same

OhmV-IR:
> A few nitpicks
> 
> * Skull name is aqua, you can add `<white>` in the item name to override this
> * 'Hyperactive' with skull doesn't really fit here. Maybe Hyper_activated_? Also the files and keys and such are still 'super' not 'hyperactive'
> * I would suggest nerfing the cost a bit, maybe 1 nether star instead of 3?
> * Hyper wither skull is IMO less useful than the reactivated one because the explosions are the same size but the head slows down quickly. Not sure how this can be fixed though

it appears that the charged skulls have this property in vanilla of deceleration, tried some stuff to get rid of the deceleration(eg entity.setAcceleration(0,0,0)) but these did nothing so I don't think there's an easy way to do so. I've buffed the initial velocity and reduced the cooldown on the hyper variant as a way to compensate for this. went with the hyper activated name, changed keys and lang to match

Seggan (approved):
We can find balance issues in playtesting

PR #115, opened 2025-08-03
Fix immobilizer thread safety issues

### LINKED ISSUE #114: Thread safety issue in immobilizer
https://discord.com/channels/1329177304857055273/1329177304857055276/1401336960332861593

OhmV-IR:
Closes #114 
At least someone noticed my mistake
ty to that random person in gen chat :heart:
Moved both runTaskLater calls from the async to the non-async version, since there is no ConcurrentHashSet both calls must be made on the main thread to avoid unintended consequences

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/building/Immobilizer.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/building/Immobilizer.java
@@ -74,11 +74,11 @@ public class Immobilizer extends PylonBlock implements PylonPiston {
             }
             frozenPlayers.add(player);
             for (int i = 0; i < duration / particlePeriod; i++) {
-                Bukkit.getScheduler().runTaskLaterAsynchronously(PylonBase.getInstance(), new PlayerVFX(player, this), (long) i * particlePeriod);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/pylonmc/pylon/base/content/building/Immobilizer.java
```diff
@@ -47,7 +48,7 @@ public Item(@NotNull ItemStack stack) {
         }
     }
 
-    public static Set<Player> frozenPlayers = new HashSet<>();
+    public static Set<Player> frozenPlayers = ConcurrentHashMap.newKeySet();
```
Seggan, line 51: Don't need it concurrent if its run sync

Seggan (changes requested):

src/main/java/io/github/pylonmc/pylon/base/content/building/Immobilizer.java
Seggan: That's my point: now that it's sync, we dont need this set to be concurrent as it will only ever be accessed from one thread

Seggan (approved):

LordIdra (approved):

PR #120, opened 2025-08-06
Add Pit Kiln

Seggan:
The Primitive Smelting Solution You Have Always Wished For!™

<sub>Pylon, Inc. does not claim any responsibility for any accident, death, burn, drinking hot gold, or immolation caused by use of the Pit Kiln</sub>


src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+        item.subtract(taken);
+    }
+
+    @Override
+    public void tick(double deltaSeconds) {
+        if (isFormedAndFullyLoaded()) {
+            for (Vector3i relative : getComponents().keySet()) {
+                BlockPosition block = new BlockPosition(getBlock()).addScalar(relative.x(), relative.y(), relative.z());
+                Waila.addWailaOverride(block, this::getComponentWaila);
+            }
```
OhmV-IR, line 143: adding waila overrides every tick? shouldn't this just be done once onFormedAndFullyLoaded

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+                        coalBlock.setType(Material.AIR);
+                    }
+                }
+            }
+        } else {
+            if (isProcessing()) {
+                processingTime = Double.NaN;
+                processing.clear();
+            }
+            removeWailas();
```
OhmV-IR, line 174: same comment here about removing wailas every tick that the multiblock isn't fully formed

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+            }
+            if (isProcessing()) {
+                processingTime -= deltaSeconds;
+                if (processingTime <= 0) {
+                    processingTime = Double.NaN;
+                    int multiplier = 2;
+                    for (Vector3i top : TOP_POSITIONS) {
+                        Block topBlock = getBlock().getRelative(top.x(), top.y(), top.z());
+                        if (topBlock.getType() != Material.PODZOL) {
+                            multiplier = 1;
```
OhmV-IR, line 155: make the multipliers settings

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+                baseKey("brass"),
+                List.of(new ItemStack(Material.COPPER_INGOT, 2), BaseItems.ZINC_INGOT),
+                List.of(BaseItems.BRASS_INGOT)
+        ));
+    }
+
+    public record Recipe(
+            @NotNull NamespacedKey key,
+            @NotNull List<ItemStack> input,
+            @NotNull List<ItemStack> output
```
OhmV-IR, line 289: add an exception in constructor that input can only be x elements large, same with output

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+            }
+
+            return Gui.normal()
+                    .setStructure(
+                            "# # # # # # # # #",
+                            "# , , # # # . . #",
+                            "# , , # p # . . #",
+                            "# , , # # # . . #",
+                            "# # # # # # # # #"
+                    )
```
OhmV-IR, line 338: setStructure seems to show 6 max inputs/outputs but the UnclickableInventory is 9 in size

OhmV-IR (changes requested):

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
Seggan: removing is fine, it's just a map remove, it has practically 0 performance impact

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+public final class PitKiln extends PylonBlock implements
+        PylonSimpleMultiblock, PylonInteractableBlock, PylonTickingBlock {
+
+    private static final int CAPACITY = Settings.get(BaseKeys.PIT_KILN).getOrThrow("capacity", Integer.class);
+    private static final int PROCESSING_TIME_SECONDS =
+            Settings.get(BaseKeys.PIT_KILN).getOrThrow("processing-time-seconds", Integer.class);
+
+    public static final class Item extends PylonItem {
+
+        private static final int CAPACITY = Settings.get(BaseKeys.PIT_KILN).getOrThrow("capacity", Integer.class);
```
LordIdra, line 52: This should probably be a property of PitKiln rather than the Item

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+        int taken = Math.min(CAPACITY - currentAmount, item.getAmount());
+        if (taken <= 0) return;
+
+        contents.merge(item.asOne(), taken, Integer::sum);
+        item.subtract(taken);
+    }
+
+    @Override
+    public void tick(double deltaSeconds) {
+        if (isFormedAndFullyLoaded()) {
```
LordIdra, line 139: Worth inverting this condition and using an early return I think

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+            for (Vector3i relative : getComponents().keySet()) {
+                BlockPosition block = new BlockPosition(getBlock()).addScalar(relative.x(), relative.y(), relative.z());
+                Waila.addWailaOverride(block, this::getComponentWaila);
+            }
+            if (!isProcessing()) {
+                tryStartProcessing();
+            }
+            if (isProcessing()) {
```
LordIdra, line 147: Probably also worth early returns here if above comment followed

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+    private WailaConfig getComponentWaila(@NotNull Player player) {
+        if (isProcessing()) {
+            return new WailaConfig(Component.translatable(
+                    "pylon.pylonbase.waila.pit_kiln",
+                    PylonArgument.of(
+                            "time",
+                            UnitFormat.formatDuration(Duration.ofSeconds((long) processingTime))
+                    )
+            ));
+        } else {
```
LordIdra, line 187: Unneeded else

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+                List.of(BaseItems.BRONZE_INGOT)
+        ));
+        Recipe.RECIPE_TYPE.addRecipe(new Recipe(
+                baseKey("brass"),
+                List.of(new ItemStack(Material.COPPER_INGOT, 2), BaseItems.ZINC_INGOT),
+                List.of(BaseItems.BRASS_INGOT)
+        ));
+    }
+
+    public record Recipe(
```
LordIdra, line 286: This should now be separate in a PitKilnRecipe file as of latest PRs (along with the above static blocks)

src/main/resources/lang/en.yml
```diff
@@ -410,6 +417,25 @@ item:
...
 
+  pedestal:
+    name: "Pedestal"
+    lore: |-
+      <arrow> <insn>Right click</insn> to set the displayed item
+      <arrow> <insn>Shift right click</insn> to rotate the item
+
+  pit_kiln:
+    name: "Pit Kiln"
+    lore: |-
```
LordIdra, line 428: Maybe an instruction that completing the multiblock starts it off, just to clarify.

src/main/resources/lang/en.yml
```diff
@@ -410,6 +417,25 @@ item:
...
+      <star> Multiblock
+      <arrow> A primitive version of the smeltery
+      <arrow> Can perform any smeltery recipe under <unit:celsius>1100</unit>
+      <arrow> <insn>Right click</insn> to insert items
+      <arrow> Using fire and soul campfire smelts items at normal speed, using normal campfire smelts items at half speed, and using soul fire smelts items at double speed
+      <arrow> Using podzol instead of dirt doubles yield
```
LordIdra, line 434: Continuing from the 'should be configurable' comments above - the amounts here should probably be something like

Using a campfire in the multiblock smelts at `0.5x speed`, fire or a soul campfire smelts at `1.0x speed`, and soul fire smelts at `1.5x speed`

Using dirt in the multiblock gives `1.0x` yield, and using podzol gives `2.0x` yield

(with the code sections in a different color)

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,415 @@
...
+                contents.merge(inputItem.asOne(), -inputItem.getAmount() * ratio, Integer::sum);
+            }
+            Set<ItemStack> outputItems = new HashSet<>(recipe.output().size());
+            for (ItemStack outputItem : recipe.output()) {
+                ItemStack outputCopy = outputItem.asOne();
+                outputCopy.setAmount(outputItem.getAmount() * ratio);
+                outputItems.add(outputCopy);
+            }
+            processing.addAll(outputItems);
+            double multiplier = switch (getBlock().getRelative(FIRE_POSITION.x(), FIRE_POSITION.y(), FIRE_POSITION.z()).getType()) {
```
LordIdra, line 228: Should be configurable

LordIdra:
> I've held off on config until we can figure out how to express that in lore, otherwise ready for re-review

I made suggestions in one of my comments in case you didn't see

Seggan:
Ah oops

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,320 @@
...
+            if (processingTime != null) {
+                processingTime = Double.NaN;
+                processing.clear();
+            }
+            return;
+        }
+        if (processingTime == null) {
+            tryStartProcessing();
+        }
+        if (processingTime != null) {
```
LordIdra, line 158: nitpick time!!! can we invert these ifs?

LordIdra (approved):

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
Seggan: As I have already stated, no because the second one may run if the first's body starts processing

Seggan: Maybe the second one yes, but not the first

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,320 @@
...
+                    };
+                    topBlock.setType(Material.COARSE_DIRT);
+                }
+                for (ItemStack outputItem : processing) {
+                    contents.merge(outputItem.asOne(), (int) Math.floor(outputItem.getAmount() * multiplier), Integer::sum);
+                }
+                processing.clear();
+                for (Vector3i coal : COAL_POSITIONS) {
+                    Block coalBlock = getBlock().getRelative(coal.x(), coal.y(), coal.z());
+                    coalBlock.setType(Material.AIR);
```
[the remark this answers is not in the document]
Seggan: And what happens if I do? I can't just *not* break the block, it's an integral part of the pit kilns operation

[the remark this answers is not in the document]
LordIdra: Just fire the event but ignore the result is my suggestion - better than not firing at all

LordIdra:
> Just to flag, if you override `checkFormed` and then use `if fullyFormedAndLoaded` afterwards, the code in that if block will be run whenever the multiblock is formed (and any code in the else block when the multiblock is unformed). We should 100% have separate methods for this though, it would be easy to implement and very useful, so I'm creating an issue for that. But either way I'd suggest moving all the multiblock checking into there and out of ticking, because it's _significantly_ better for performance and makes the logic much more nicely separated.

Just want to make sure you saw this

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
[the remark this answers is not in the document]
Seggan: `BlockBreakEvent` requires a player

Seggan:
Uhh guys this may be a bit too long
image

LordIdra:
> Uhh guys this may be a bit too long

We can put it in the guide when that's a thing

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
```diff
@@ -0,0 +1,324 @@
...
+            contents.merge(
+                    outputItem.asOne(),
+                    (int) Math.floor(outputItem.getAmount() * multiplier),
+                    Integer::sum
+            );
+        }
+        processing.clear();
+        for (Vector3i coal : COAL_POSITIONS) {
+            Block coalBlock = getBlock().getRelative(coal.x(), coal.y(), coal.z());
+            coalBlock.setType(Material.AIR);
```
OhmV-IR, line 184: While I hate not calling a block break evt here, I don't think you can do it because what if the player dcs before it finishes? Worse than not calling an event is doing it inconsistently.

OhmV-IR (commented):
Looks fine cw

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/PitKiln.java
Seggan: I cannot call the event *at all* because it requires a player param

Issue #113, opened 2025-07-22
smeltery controller places the displays before smeltery multiblock is complete, which looks strange


LordIdra:
<img width="1234" height="590" alt="Image" src="https://github.com/user-attachments/assets/e6d663fc-927e-4570-8391-0ac0c16ad654" />

LordIdra:
closed by https://github.com/pylonmc/pylon-core/pull/206

Issue #127, opened 2025-08-09
it seems that missing %allowed-temperature%


haiman233:
<img width="2302" height="1156" alt="Image" src="https://github.com/user-attachments/assets/4232d17b-9988-478e-9552-65fcb67f55d6" />

PR #130, opened 2025-08-09
Orbital bombardment of bugs

LordIdra:
closes https://github.com/pylonmc/pylon-base/issues/113

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryController.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryController.java
@@ -305,6 +305,17 @@ public final class SmelteryController extends SmelteryComponent
 
     @Override
     public boolean checkFormed() {
+        boolean formed = isFormed();
+
+        // set pixels invisible if multiblock not formed, and visible if multiblock formed
+        for (SimpleTextDisplay display : getPixels()) {
+            display.getEntity().text(Component.text(formed ? " " : ""));
+        }
+
+        return formed;
+    }
+
+    private boolean isFormed() {
         double previousCapacity = capacity;
         height = 0;
         capacity = 0;


Seggan (approved):

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/MixingPot.java
```diff
@@ -115,12 +115,18 @@ private void updateCauldron() {
 
     @Override
     public void onInteract(@NotNull PlayerInteractEvent event) {
-        // Only allow inserting water - events trying to insert lava will be cancelled
-        if (event.getItem() != null && Set.of(Material.BUCKET, Material.WATER_BUCKET, Material.GLASS_BOTTLE).contains(event.getMaterial())) {
+        // Don't allow fluid to be manually inserted/removed
```
Seggan, line 118: Why not? Could be useful with water

Seggan (approved):

OhmV-IR (changes requested):
Actually fixed bugs:
#113 
#138 
#127 
https://github.com/pylonmc/pylon-core/issues/182 (no idea what the "wrong way" was, but it looks fine ig)
Not actually fixed:
https://github.com/pylonmc/pylon-core/issues/187
#144

LordIdra:
Both seem to be fixed..? The watering can issue persists (sort of) if you're in creative, but I'm definitely cancelling the PlayerBucketFillEvent so I don't think this is our problem. Not worth the time to investigate IMO.

OhmV-IR (changes requested):
The ferroduralum tools have hardcoded values for durability, should be item settings, but I think we should just yeet ferroduralum at this point since it doesn't integrate with the smelting system

LordIdra:
> The ferroduralum tools have hardcoded values for durability, should be item settings, but I think we should just yeet ferroduralum at this point since it doesn't integrate with the smelting system

This was already the case before so decided not to do add in PR. progression PR changes ferroduralum into bronze

PR #148, opened 2025-08-11
Actually fix lumber axe

### LINKED ISSUE #143: The lumber axe crashed the server，when logging.
[log.txt](https://github.com/user-attachments/files/21703848/log.txt)

OhmV-IR:
Closes #143 @LordIdra we could keep the recursion limit too as an additional safety measure if you want, but if you want to get rid of it you can go ahead

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/tools/LumberAxe.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/tools/LumberAxe.java
@@ -28,7 +28,7 @@ public class LumberAxe extends PylonItem implements PylonTool {
         super(stack);
     }
 
-    private final Set<Event> eventsToIgnore = HashSet.newHashSet(0);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/pylonmc/pylon/base/content/tools/LumberAxe.java
```diff
@@ -28,7 +28,7 @@ public LumberAxe( @NotNull ItemStack stack) {
         super(stack);
     }
 
-    private final Set<Event> eventsToIgnore = HashSet.newHashSet(0);
+    private static final Set<Event> eventsToIgnore = HashSet.newHashSet(0);
```
Seggan, line 31: This may leak memory if the events are never removed. Also why not `new HashSet><>()`?

LordIdra (commented):
Nice. I did notice one other issue which might be good to add to this PR - it seems you can't break any non-log blocks with the lumber axe

Issue #152, opened 2025-08-17
Divided by 0


balugaq:
[16:28:40 ERROR]: Could not pass event PlayerInteractEvent to PylonCore v0.9.0
java.lang.IllegalStateException: Cannot format NaN or infinite values
        at pylon-core-0.9.0.jar/io.github.pylonmc.pylon.core.util.gui.unit.UnitFormat.format(UnitFormat.kt:53) ~[pylon-core-0.9.0.jar:?]
        at pylon-base-MODIFIED.jar/io.github.pylonmc.pylon.base.content.machines.smelting.SmelteryController$InfoItem.getItemProvider(SmelteryController.java:208) ~[pylon-base-MODIFIED.jar:?]
        at xyz.xenondevs.invui.item.Item.getItemProvider(Item.java:36) ~[?:?]
        at xyz.xenondevs.invui.gui.SlotElement$ItemSlotElement.getItemStack(SlotElement.java:44) ~[?:?]
        at xyz.xenondevs.invui.window.AbstractWindow.redrawItem(AbstractWindow.java:109) ~[?:?]
        at xyz.xenondevs.invui.window.AbstractSingleWindow.initItems(AbstractSingleWindow.java:58) ~[?:?]
        at xyz.xenondevs.invui.window.AbstractWindow.open(AbstractWindow.java:279) ~[?:?]
        at pylon-core-0.9.0.jar/io.github.pylonmc.pylon.core.block.base.PylonGuiBlock.onInteract(PylonGuiBlock.kt:59) ~[pylon-core-0.9.0.jar:?]
        at pylon-core-0.9.0.jar/io.github.pylonmc.pylon.core.block.BlockListener.onPlayerBlockInteract(BlockListener.kt:496) ~[pylon-core-0.9.0.jar:?]
        at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:80) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
        at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:70) ~[paper-api-1.21.4-R0.1-SNAPSHOT.jar:?]
        at io.papermc.paper.plugin.manager.PaperEventManager.callEvent(PaperEventManager.java:54) ~[paper-1.21.4.jar:1.21.4-232-12d8fe0]
[...]

Issue #128, opened 2025-08-09
Unstable smeltery temperature simulation with a 2 block high smeltery and many burners


haiman233:
<img width="1956" height="1167" alt="Image" src="https://github.com/user-attachments/assets/00716961-174f-473d-97b5-2d2b2e4a8718" />

https://github.com/user-attachments/assets/079a9acb-bcf4-41c8-a7ca-85a97da68354

LordIdra:
Leaving this to @Seggan because the smeltery code is still an enigma to me

haiman233:
When I try to increase the number of smeltery_burner while the smelting furnace is running, and put in carbon dust with higher energy, I will get the following error. If I rearrange the smelting furnace controller, I will get a bug with negative temperature.

haiman233:
09:13:38 错误]: [PylonCore] An error occurred while ticking block -28, 104, -140 in world of type pylonbase:smeltery_controller
[09:13:38 警告]: java.lang.IllegalArgumentException: Red[-4] is not between 0-255
[09:13:38 警告]: 	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:191)
[09:13:38 警告]: 	at org.bukkit.Color.<init>(Color.java:205)
[09:13:38 警告]: 	at org.bukkit.Color.fromRGB(Color.java:140)
[09:13:38 警告]: 	at pylon-base-0.2.0.jar//io.github.pylonmc.pylon.base.util.HslColor.toRgb(HslColor.java:55)
[09:13:38 警告]: 	at pylon-base-0.2.0.jar//io.github.pylonmc.pylon.base.content.machines.smelting.SmelteryController.updateFluidDisplay(SmelteryController.java:559)
[09:13:38 警告]: 	at pylon-base-0.2.0.jar//io.github.pylonmc.pylon.base.content.machines.smelting.SmelteryController.tick(SmelteryController.java:613)
[09:13:38 警告]: 	at pylon-core-SNAPSHOT.jar//io.github.pylonmc.pylon.core.block.TickManager$startTicker$1.invokeSuspend(TickManager.kt:72)
[09:13:38 警告]: 	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
[09:13:38 警告]: 	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
[...]

PR #153, opened 2025-08-17
Fix negative temperature

balugaq:
Desmos simulation: https://www.desmos.com/calculator/2jsf3qwyb5

Origin formula function graph like:
45c89ebf2f614605ff0edc9ea9689980

When blue > 1382, green < -973, in the next term, blue will be -973, and green will be less than -973, over and over again. It causes smeltery's temperature becomes lower accidentally.

If we remove `radiativeHeatLoss`, the function graph will be fine like:
0fc2f79ccc731e274c50c1ce5bab8a83

Finally the temperature will be stopped at 20 (Room temperature)


balugaq:
Closes #128 
Closes #152

balugaq:
yeah, but we can use other function to fix it,it is like add a `ln` in calculation, or other functions

balugaq:
but i m not good at math/physics, so that is the easiest way to fix it. I wonder what function can both fix negative and keep it logarithmic

Issue #170, opened 2025-08-25
Ferroduralum axe can't strip woods


balugaq:
Same with lumber axe

balugaq:
shovel and hoe also can't interact with grass block

PR #191, opened 2025-08-28
Convert bronze items to use configs and add bronze armour

OhmV-IR:
While there's no armor data component yet, the attribute modifiers work for what we need to do
converted the bronze items to use item settings


Seggan:
Armor data component should exist, we in 1.21.8 by now

OhmV-IR:
> Armor data component should exist, we in 1.21.8 by now

double checked the javadocs, nothing

OhmV-IR:
note that the bronze hoe is broken due to #170 atm

LordIdra (approved):

PR #195, opened 2025-08-30
Fix #181

### LINKED ISSUE #181: Infinite fluid in Hydraulic Table Saw
Image

It needs 400 mB hydraulic fluid and works forever, while no dirty hydraulic fluid was produced

What caused it:

Missing `addFluid` and `removeFluid`
Image

balugaq:
Close #181


LordIdra (changes requested):
Your changes ignore the tick interval. If TICK_INTERVAL is 1, the consumption will be 20x higher than it should be

LordIdra (changes requested):

Seggan (approved):
Besides Idra comments LGTM

PR #222, opened 2025-09-17
Fix core #273

Intybyte:
Fix core #273

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/science/Loupe.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/science/Loupe.java
@@ -65,13 +65,11 @@ public class Loupe extends PylonItem implements PylonInteractor, PylonConsumable
         if (offhand.isEmpty()) {
             player.sendMessage(Component.translatable("pylon.pylonbase.message.loupe.nothing"));
             event.setCancelled(true);
-            player.setCooldown(getStack(), 20);
             return;
         }
         if (PylonItem.fromStack(offhand) != null) {
             player.sendMessage(Component.translatable("pylon.pylonbase.message.loupe.is_pylon"));
             event.setCancelled(true);
-            player.setCooldown(getStack(), 20);
             return;
         }
         ItemRarity rarity = offhand.getType().getDefaultData(DataComponentTypes.RARITY);
@@ -81,7 +79,6 @@ public class Loupe extends PylonItem implements PylonInteractor, PylonConsumable
         if (items.getOrDefault(offhand.getType(), 0) >= maxUses) {
             player.sendMessage(Component.translatable("pylon.pylonbase.message.loupe.already_examined"));
             event.setCancelled(true);
-            player.setCooldown(getStack(), 20);
         }
     }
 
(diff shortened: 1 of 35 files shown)


LordIdra:
I think you need to remove the eg `player.setCooldown(getStack(), 20);` code in the items that the cooldown has been added to

Intybyte:
now fixes #224
allows for scanning blocks and items

src/main/java/io/github/pylonmc/pylon/base/content/science/Loupe.java
LordIdra, line 1: Loupe lore needs to be updated with the new behaviour

LordIdra (changes requested):
I'm not sure if this is the best solution to the loupe issue as it's a slight workaround and adds more behaviour to the loupe which we now have to maintain and document. Is it not possible to simply set up an item interact listener which cancels any interactions in the offhand if the main hand contains a loupe?

src/main/java/io/github/pylonmc/pylon/base/content/science/Loupe.java
LordIdra: Oh wait yes, you couldn't really cancel the events because the clientside prediction would lose its shit. Though I'd suggest if we're going to allow scanning items on the floor, that should be the only way to do it rather than also being able to hold items in the offhand.

Intybyte: Yeah that is why I wanted ground scanning, even cancelling event it would still play with it client side lol, the loupe seems to behave like thaumcraft's lens back in the day (not sure if it has been changed) and I basically used the same logic, also no real reason to remove the offhand mode

LordIdra: Well there's just no reason to have two ways to do the exact same thing. It's more code, more text the user won't bother to read, etc

LordIdra: > ngl I prefer the current way the loupe works

Have you seen https://github.com/pylonmc/pylon-base/issues/224

Seggan: Ah. hmm. I'll let yall figure this one out then

src/main/java/io/github/pylonmc/pylon/base/content/science/Loupe.java
```diff
@@ -62,58 +67,181 @@ public Loupe(@NotNull ItemStack stack) {
...
+            // found valid item that hasn't been scanned yet
+            if (items.getOrDefault(stack.getType(), 0) < maxUses) {
+                return item;
+            }
+        }
+
+        return null;
+    }
+
+    private static String normalizeName(String original) {
```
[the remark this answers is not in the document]
LordIdra: I don't think this should be needed. Can we create a new itemstack from the material and get the name from there? We do something similar in parts of the codebase but can't remember where exactly - maybe caster?

LordIdra:
I think message as fine, because we should generally avoid action bar / boss bar as many plugins take them over for various things. (aurelium skills for example)

Intybyte:
Figured out the reason brick mold didn't active the coded in cooldown, it was because it wasn't a consumable so minecraft didn't detect it, so for non consumable the cooldown must be coded in

PR #230, opened 2025-09-18
Add Hydraulic excavator

LordIdra:


LordIdra:
> LGTM CW & IG, one nice thing would be some sort of animation, or at the very least an indication that it started, but won't block because of that

Yeah I was thinking about this but not really sure what to add. I think in future I will (re-)add a WAILA display to hydraulic machines with their status. For now it's fine because it should be obvious when it's stopped I think

Seggan:
Maybe like a particle going from the block to the mined block?

Seggan (approved):

PR #259, opened 2025-09-23
Updated Loupe Examining

JustAHuman-xD:
Switches to use a raytrace for finding the targeted block/entity, feels better as a player as it now respects exactly what you're targeting.
Redoes the logic for starting & ending scanning to give more feedback to the player, add some polish, fix a few edge cases, and respect event results.
It also now allows you to use the loupe without examining something, so players can use it as a sort of magnifying glass or just have fun with it like you would a spy glass.

Current "Issues":
- Scanning a double tall blocks top, will still drop the block in cases like doors
- Scanning something with an inventory will void its contents

Todo:
- Entity Examining
- Potentially some sort of index of Examined things?
  - Don't want it to turn into just looking through pages to find what you need to scan
  - But having some index helps fuel completionist players
  - Potentially show only what you have examined & #/Posible


LordIdra:
Just want to check, is this ready for review or is it intentional that it's still a draft?

Intybyte:
> > Just want to check, is this ready for review or is it intentional that it's still a draft?
> 
> Its still a draft because I haven't added the entity examining or index yet, namely because we haven't really discussed either enough for me to start on an implementation

tbh i am for replacing the loupe pdc entry from a material -> integer to a key -> integer, and also adding proper events for loupe so addons can add their custom research behaviour if they want to, as the loupe is pretty much a key item in pylon

LordIdra:
yolo

PR #266, opened 2025-09-26
Fix hardcoded item settings

### LINKED ISSUE #262: Some pylon items have hardcoded settings
Should be replaced with `PylonItem#getSettings()`

Intybyte:
Fixes #262

Not all can be replaced, like classes that have static inner classes which reference static fields can't be unharcoded, i propose making such classes final, one example is HydraulicGrindstoneTurner

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryInputHatch.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryInputHatch.java
@@ -1,10 +1,8 @@
 package io.github.pylonmc.pylon.base.content.machines.smelting;
 
-import io.github.pylonmc.pylon.base.BaseKeys;
 import io.github.pylonmc.pylon.core.block.base.PylonEntityHolderBlock;
 import io.github.pylonmc.pylon.core.block.base.PylonFluidBlock;
 import io.github.pylonmc.pylon.core.block.context.BlockCreateContext;
-import io.github.pylonmc.pylon.core.config.Settings;
 import io.github.pylonmc.pylon.core.config.adapter.ConfigAdapter;
 import io.github.pylonmc.pylon.core.content.fluid.FluidPointInteraction;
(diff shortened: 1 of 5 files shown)


LordIdra:
Regarding classes referencing fields in the item class - we have so far just been duplicating the relevant item settings inside the item. I don't think there's really a better solution than this unfortunately

Seggan (approved):

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryInputHatch.java
```diff
@@ -15,8 +13,8 @@
 import org.bukkit.persistence.PersistentDataContainer;
 import org.jetbrains.annotations.NotNull;
 
-public final class SmelteryInputHatch extends SmelteryComponent implements PylonFluidBlock, PylonEntityHolderBlock {
```
OhmV-IR, line 18: Why are we removing final from this class?

OhmV-IR (approved):

LordIdra:
> > Regarding classes referencing fields in the item class - we have so far just been duplicating the relevant item settings inside the item. I don't think there's really a better solution than this unfortunately
> 
> Should I duplicate entries or mark those item classes as final? Both are valid options imo

Marking item classes as final does not fix the issue

Intybyte:
> > > Regarding classes referencing fields in the item class - we have so far just been duplicating the relevant item settings inside the item. I don't think there's really a better solution than this unfortunately
> > 
> > 
> > Should I duplicate entries or mark those item classes as final? Both are valid options imo
> 
> Marking item classes as final does not fix the issue

It does, the whole issues is that if people extend classes they can't modify their behaviour with their key, by making them final the addon dev that might want to extend one of our classes can't do it and he doesn't have to troubleshoot why the configs don't work, or at least that was my idea

OhmV-IR:
has merge conflicts, needs rebase, skipping over for my review sesh today, ping again when it's fixed

LordIdra:
happy to merge and do that in another pr then

PR #268, opened 2025-09-27
Use vanilla entities in pylon entity block holder

LordIdra:
Companion to https://github.com/pylonmc/pylon-core/pull/341

Converts all blocks to use display entities instead of Pylon entities wrapping display entities.

Adds model data to all (relevant) item displays so they can be textured,


LordIdra:
> Smeltery burner and controller might also want a "active" key or similar (see https://discord.com/channels/1329177304857055273/1421566978040266862/1422242081694289971)

Out of scope of this PR I believe as it's only really touching entities

PR #297, opened 2025-10-03
Fix/operation clay ball

Intybyte:
Fixed stuff

- Iron ingots
- Dyes

To fix

- Potion materials

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/BaseItems.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/BaseItems.java
@@ -175,7 +175,8 @@ public final class BaseItems {
         BasePages.RESOURCES.addItem(RAW_TIN);
     }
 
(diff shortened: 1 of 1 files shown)


LordIdra:
> Why does sugar, glowstone dust, gunpowder need to be a clay ball?

Oh I assume because of potion materials. It should already be fixed by listeners but it prevents them from being inserted into a brewing stand on the client ig

Seggan (changes requested):
The problem of tools should be solved by properly implementing anvil repair for custom tools

Intybyte:
> The problem of tools should be solved by properly implementing anvil repair for custom tools

The main problem is golden tools (so bronze tools) getting picked up by piglin, I didn't change other tools, the actual event logic for such cases was handled in another PR by justin iirc

If we need to make them repairable there is a component for that also, but only allows itemtypes and not itemstacks so when necessary we need to hook to an event, however it is out of scope of this PR

LordIdra (approved):
Happy with code, not tested ingame

PR #298, opened 2025-10-03
Fix vacuum hopper drops

### LINKED ISSUE #237: Vacuum hopper does not drop items when broken

Intybyte:
Fixes #237


LordIdra (approved):
Have not tested, seems like a simple enough change.

PR #319, opened 2025-10-04
[ref:fix-build] Migrate drilling and molding recipes

LordIdra:
Change drilling recipe to use config files. This removes the circular dependency in BaseItems between core chunks and core drills. The recipe is still purely for display. I have added a comment in the file to indicate this.

Change molding recipe to also use config files and changed molding mechanic to actually respect the recipe. This allows us to also migrate this to config files instead of having it in BaseItems.


JustAHuman-xD (approved):

OhmV-IR (changes requested):
Clicking on core chunks in guide doesn't show any recipes for how to produce them,

LordIdra:
> Clicking on core chunks in guide doesn't show any recipes for how to produce them,

If everything else is fine I'll just merge this and open an issue to fix this later, as I want to fix master and get a release out rn

OhmV-IR:
> > Clicking on core chunks in guide doesn't show any recipes for how to produce them,
> 
> If everything else is fine I'll just merge this and open an issue to fix this later, as I want to fix master and get a release out rn

yeah that's the only thing, opened the issue and changed my review to approval

OhmV-IR (approved):

PR #348, opened 2025-10-07
Remove XML Bypass (Fix Publishing)

JustAHuman-xD:


Seggan (approved):
No idea how the problem randomly fixed itself but ok

PR #349, opened 2025-10-08
Add bloomery

Seggan:
Also make cobalt and nickel temporarily unobtainable.

Bloomery is basically the one in the vid, except with better particles.


LordIdra:
(Ready for re-review?)

LordIdra (approved):
Nice work, I think the changes are a big improvement

Issue #338, opened 2025-10-05
Farmer talisman


LordIdra:
Get more drops from crops. Maybe like 10% more t1, 20% more t2, 30% more t3

Issue #337, opened 2025-10-05
Bartering talisman


LordIdra:
Increases luck when bartering. not sure how this would be implemented.

Issue #335, opened 2025-10-05
Talisman ring


LordIdra:
A ring which holds talismans. Only 1 can be used at a time.

Could add multiple tiers, each with more slots - this could be cool and maybe you would need materials from each progression stage to upgrade the ring.

Why?
- Reduces inventory clutter. If we're going to be adding lots of talismans, this is important
- Performance. Lots of talismans = lots of comparisons in inventory tickers. This will reduce the number of comparisons to 1

Issue #334, opened 2025-10-05
Mob farmer talisman


LordIdra:
Lowers mob breeding cooldown when you breed a mob. Maybe like 0.9x t1, 0.8x t2, 0.7x t3

PR #366, opened 2025-10-13
Add talismans

OhmV-IR:
Implements the following:
- [x] Mob farmer talisman #334 
- [x] Hunger talisman #333 
- [x] Luck talisman #332 
- [x] Hunting talisman #331 
- [x] Enchanting talisman #330 
- [x] Bartering talisman #337 
- [x] Water breathing talisman #329 
- [x] Farmer talisman #338 
- [x] Recipes for talismans
Makes tick interval of talismans configurable. Leaving #335 for future pr since with the Talisman class, this likely won't need to be refactored much to accommodate that.


src/main/java/io/github/pylonmc/pylon/base/content/talismans/base/Talisman.java
```diff
@@ -0,0 +1,85 @@
...
+            removeEffect(player);
+            applyEffect(player);
+        }
+    }
+
+    @Override
+    public void onRemovedFromInventory(@NotNull Player player) {
+        PylonInventoryEffectItem.super.onRemovedFromInventory(player);
+        Integer currentTalismanLevel = player.getPersistentDataContainer().get(getTalismanKey(), PersistentDataType.INTEGER);
+        if(currentTalismanLevel == null) return; // really shouldn't happen, but in this case less likely to crash by not calling removeEffect
```
Seggan: Formatting

src/main/java/io/github/pylonmc/pylon/base/content/talismans/BarteringTalisman.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+
+public class BarteringTalisman extends PDCKeyTalisman<Float, Float> {
+    public static final NamespacedKey BARTERING_TALISMAN_KEY = new NamespacedKey(PylonBase.getInstance(), "bartering_talisman");
+    public static final NamespacedKey BARTERING_TALISMAN_NO_CONSUME_KEY = new NamespacedKey(PylonBase.getInstance(), "bartering_talisman_no_consume_chance");
+    public final float chanceToNotConsumeInput = getSettings().getOrThrow("chance-to-not-consume-input", ConfigAdapter.FLOAT);
+    private static final Random RNG = new Random();
```
Seggan: use ThreadLocalRandom instead

src/main/java/io/github/pylonmc/pylon/base/content/talismans/EnchantingTalisman.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public static class EnchantingListener implements Listener {
+        @EventHandler
+        public void onPreEnchant(PrepareItemEnchantEvent event) {
+            if (!event.getEnchanter().getPersistentDataContainer().has(ENCHANTING_TALISMAN_BONUS_KEY)) {
+                return;
+            }
+            @SuppressWarnings("DataFlowIssue")
+            double bonusLevelChance = event.getEnchanter().getPersistentDataContainer().get(ENCHANTING_TALISMAN_BONUS_KEY, PersistentDataType.DOUBLE);
```
Seggan:
```suggestion
            double bonusLevelChance = event.getEnchanter().getPersistentDataContainer().get(ENCHANTING_TALISMAN_BONUS_KEY, PersistentDataType.DOUBLE);
            if (bonusLevelChance == null) {
                return;
            }
```

Seggan (changes requested):

LordIdra:
(Let me know when this is ready for re-review - not sure if it is rn)

src/main/java/io/github/pylonmc/pylon/base/content/talismans/EnchantingTalisman.java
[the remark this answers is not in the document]
OhmV-IR: I've gotten it to work. Problem is that since we're using the enchanting table seed for the rng seed of the talisman, it will only up the same enchantments over and over again, in my case sharpness 1 -> 2 and bane of arthropods 1 -> 2. Fixed by assigning each item that comes through an enchanting table with the talisman a uuid. The exploit I could see wrt to that is that someone nefarious could run thousands of items through the enchanting table and balloon the file size of the world. Although at 16 bytes/uuid, it would take 67,108,864 items to fill up 1GB of space which is a lot so I'm not insanely worried. It's still not perfect though, if there are 3 sharpness 1s in the same view, they will all have their levels increased by 1. This is because the chance has to be deterministic from the etable, item id, and enchant type in order for us to be able to apply the enchant in the EnchantItemEvent since we don't have access to the EnchantmentOffer order.

src/main/java/io/github/pylonmc/pylon/base/content/talismans/BarteringTalisman.java
```diff
@@ -0,0 +1,77 @@
...
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.persistence.PersistentDataType;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class BarteringTalisman extends PDCKeyTalisman<Float, Float> {
+    public static final NamespacedKey BARTERING_TALISMAN_KEY = new NamespacedKey(PylonBase.getInstance(), "bartering_talisman");
```
LordIdra: Just FYI - you can use BaseUtils.baseKey("bartering_talisman") here. Won't comment on the other instances; up to you if you change this, it's not a big deal

src/main/java/io/github/pylonmc/pylon/base/BasePages.java
```diff
@@ -56,5 +58,6 @@ public boolean shouldDisplay(@NotNull Player player) {
 
         PylonGuide.getRootPage().addPage(BUILDING);
         PylonGuide.getRootPage().addPage(CREATIVE_ITEMS);
+        PylonGuide.getRootPage().addPage(TALISMANS);
```
LordIdra: Would make more sense to have this after 'combat' in the guide

src/main/java/io/github/pylonmc/pylon/base/content/talismans/base/AttributeTalisman.java
LordIdra: Just a little convention thing, but we don't really use `base` folders - I'd suggest just moving the folders in `content/base` to `content/talismans`

src/main/java/io/github/pylonmc/pylon/base/content/talismans/EnchantingTalisman.java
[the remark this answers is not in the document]
LordIdra: There are _much_ easier ways to inflate file size so I would not worry about that lol. I don't quite understand what's going on here but happy as long as it works

src/main/resources/lang/en.yml
LordIdra: General UI thing, might be better to separate out the whole 'while your inventory' part, maybe something like
```
<arrow> Gain saturation and hunger every %period%
<arrow> Only works while in your inventory
```

src/main/resources/lang/en.yml
```diff
@@ -1441,6 +1441,153 @@ item:
       <arrow> Inserts items into chests, furnaces, machines, and other compatible blocks
       <arrow> <attr>Transfer rate:</attr> %transfer-rate%
 
+  hunger_talisman_simple:
+    name: "Simple Hunger Talisman"
+    lore: |-
+      <arrow> Gain saturation and hunger every %period% while in your inventory
```
LordIdra: 'Gain' -> 'Increases your'

src/main/resources/lang/en.yml
```diff
@@ -1441,6 +1441,153 @@ item:
...
+    name: "Ultimate Hunger Talisman"
+    lore: |-
+      <arrow> Gain saturation and hunger every %period% while in your inventory
+      <arrow> <attr>Saturation increase:</attr> %saturation_increase%
+      <arrow> <attr>Hunger increase:</attr> %hunger_increase%
+
+  farmer_talisman_simple:
+    name: "Simple Farming Talisman"
+    lore: |-
+      <arrow> Gain a small chance to get an extra crop drop while in your inventory
```
LordIdra: Maybe reword to 'Gives you a chance to get an extra drop when breaking a crop' for slightly better clarity

src/main/resources/lang/en.yml
LordIdra: Another little UI thing: maybe instead of having stuff like 'Change for extra item', 'Chance to not consume gold' etc you could just write 'Success rate'. Would be a bit more clear and consistent and might allow you to simplify some of the talisman descriptions too as you wouldn't need to mention 'Gain a chance to...'

src/main/resources/lang/en.yml
```diff
@@ -1441,6 +1441,153 @@ item:
...
+  luck_talisman_simple:
+    name: "Simple Luck Talisman"
+    lore: |-
+      <arrow> Makes you a much luckier ducky while in your inventory
+      <arrow> <attr>Bonus luck:</attr> %bonus_luck%
+
+  luck_talisman_advanced:
+    name: "Advanced Luck Talisman"
+    lore: |-
+      <arrow> Makes you a much luckier ducky while in your inventory
```
LordIdra: Needs clarification on what the heck this means in practice

src/main/java/io/github/pylonmc/pylon/base/content/talismans/base/PDCKeyTalisman.java
```diff
@@ -0,0 +1,30 @@
+package io.github.pylonmc.pylon.base.content.talismans.base;
+
+import org.bukkit.NamespacedKey;
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.persistence.PersistentDataType;
+import org.jetbrains.annotations.NotNull;
+
+public abstract class PDCKeyTalisman<P,C> extends Talisman {
```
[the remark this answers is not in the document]
OhmV-IR: ya +62 additions -112 removals to delete it, better without it

src/main/resources/lang/en.yml
LordIdra: 'only works when in your inventory' should go before any attributes (by convention)

src/main/resources/lang/en.yml
```diff
@@ -1441,6 +1444,177 @@ item:
...
+      <arrow> <attr>Chance for extra item:</attr> %bonus_item_chance%
+      <arrow> Only works while in your inventory
+
+  hunting_talisman_ultimate:
+    name: "Ultimate Hunting Talisman"
+    lore: |-
+      <arrow> Grants you a chance to get an extra item for each drop when you kill a mob
+      <arrow> <attr>Chance for extra item:</attr> %bonus_item_chance%
+      <arrow> Only works while in your inventory
+
```
LordIdra: These 3 have yellow names (again just need`<white>` before the name)

LordIdra (approved):
nice work, just 3 last formatting things

PR #395, opened 2025-10-18
Fix Watering Can use in dispenser

### LINKED ISSUE #394: Watering Can can be used to collect water and turn into water bucket in dispenser
Image

Intybyte:
Fixes #394


LordIdra (approved):
Untested but I assume you've checked it works and code is simple, so happy to merge

PR #438, opened 2025-11-10
Fix vacuum hoppers' range being too far because of corner-case

### LINKED ISSUE #372: Vacuum hopper I range is 3 blocks when the item is north/west from it
I don't know how i'm finding these
Image

glomdom:
Resolves #372

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/VacuumHopper.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/VacuumHopper.java
@@ -75,8 +75,7 @@ public class VacuumHopper extends PylonBlock implements PylonTickingBlock, Pylon
             return; // don't vacuum if powered
         }
 
-        Location location = getBlock().getLocation().add(0.5, 0.5, 0.5);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/VacuumHopper.java
```diff
@@ -74,7 +75,8 @@ public void tick(double deltaSeconds) {
             return; // don't vacuum if powered
         }
 
-        for (Entity entity : getBlock().getLocation().getNearbyEntities(radius + 0.5, radius + 0.5, radius + 0.5)) {
+        Location location = getBlock().getLocation().add(0.5, 0.5, 0.5);
+        for (Entity entity : location.getNearbyEntities(radius + 0.5, radius + 0.5, radius + 0.5)) {
```
OhmV-IR:
```suggestion
        for (Entity entity : getBlock().getLocation().toCenterLocation().getNearbyEntities(radius + 0.5, radius + 0.5, radius + 0.5)) {
```
makes intent clearer, and avoids using our own math

OhmV-IR (changes requested):
lgtm ig but lil cw thing, but I do want it before merge

PR #439, opened 2025-11-10
Fix smeltery caster voiding items if its output is full

### LINKED ISSUE #360: Smeltery Caster voids Inputs
If the caster output slot is full it will still craft and will just void the input

glomdom:
Resolves #360

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryCaster.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryCaster.java
@@ -120,12 +120,14 @@ public final class SmelteryCaster extends SmelteryComponent implements PylonGuiB
             }
 
             ItemStack result = recipe.result();
-            int leftover = inventory.addItem(null, result);
-
-            if (leftover == 0) {
-                controller.removeFluid(bottomFluid, recipe.input().amountMillibuckets());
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/SmelteryCaster.java
```diff
@@ -119,10 +120,12 @@ public void handleClick(@NotNull ClickType clickType, @NotNull Player player, @N
...
             ItemStack result = recipe.result();
-            inventory.addItem(null, result);
-            controller.removeFluid(bottomFluid, recipe.input().amountMillibuckets());
+            int leftover = inventory.addItem(null, result);
 
-            new PylonCraftEvent<>(CastingRecipe.RECIPE_TYPE, recipe, controller).callEvent();
+            if (leftover == 0) {
+                controller.removeFluid(bottomFluid, recipe.input().amountMillibuckets());
+                new PylonCraftEvent<>(CastingRecipe.RECIPE_TYPE, recipe, controller).callEvent();
+            }
```
OhmV-IR: if there's any leftover, won't this just create the items without removing the fluid for them?
eg lets say the inventory is full and I have space for 1 item, but the recipe produces 4. 1 is placed into the inventory and 3 are left over. leftover != 0, so no fluid is removed, I take out the 1 item produced and repeat for free stuff forever. you need to add an else removing the items you added if the whole recipe couldn't be crafted.

Issue #299, opened 2025-10-03
Add a way to obtain lava more sustainably


LordIdra:
crucible maybe?

PR #449, opened 2025-11-14
[ref:vaan/feature/crucible] Crucible

Intybyte:
Fixes https://github.com/pylonmc/pylon-base/issues/299

### DIFF
--- a/src/main/resources/lang/en.yml
+++ b/src/main/resources/lang/en.yml
@@ -579,7 +579,7 @@ item:
       <arrow> Place items in the crucible using <insn>right click</insn>
       <arrow> Can hold only one item type at a time
       <arrow> The items will melt depending on the heat produced by the block below the cauldron
-      <arrow> Requires a fire underneath to work
+      <arrow> Requires a heat source underneath to work
     waila: "Crucible %item_info% %liquid_info%"
 
   monster_jerky:
@@ -1488,10 +1488,10 @@ waila:
     filled: "| %bar%/%capacity% of %fluid%"
   crucible:
     item:
-      empty: "| Empty |"
+      empty: "| No items |"
       stored: "| x%amount% %type% |"
     liquid:
-      empty: "Empty"
+      empty: "No fluid"
       filled: "%bar% (%fluid%)"
   fluid_strainer:
     straining: " | Straining %item% | %bars%"


Seggan:
Instead of adding a whole new machine

Seggan:
> That feels like shooting ducks with a cannon tho

~~Ain't no one gonna stop me~~

src/main/resources/recipes/pylonbase/crucible.yml
Seggan: Suggestion: use `#pyloncore:rocks`

LordIdra (changes requested):
This is almost identical to the mixing pot as it stands.

I would suggest instead to do what 'modded' crucibles do, where items that get thrown in are consumed (one by one) and then slowly melted. So you throw cobblestone in, and over 25 seconds, 4 mb of lava are added per second. Then, it's ready to consume the next cobblestone you throw in. (And also produces smoke gradually as it goes).

This makes balancing much easier and differentiates this enough to make it worth having alongside the mixing pot

LordIdra:
Not building with latest core (do I need to use a specific branch?)
image

LordIdra (commented):
Sorry, have realised I forgot to review this in my last PR crusade

Nice work on this, it looks better now. A couple of general things
- The fluid being displayed as water when you're melting lava is really awkward. I don't know if this PR is the place to address it given it is also a mixing pot issue as well but thought I'd mention it in case. (Open an issue if you don't want to address it here)
- Fire / soul fire does not display correctly in the guide (again not sure if this is an issue for this PR, open an issue if not)
- The particles are afaict the exact same as the mixing pot. I don't think they really suit the crucible and it feels a bit unpolished.
- I would probably avoid showing the items that can be used to heat the crucible. We can add a separate page for this when the disambiguation system is in.
- It would be great to have some way to see what is inside the crucible waiting to be melted. There's no way to know ATM. Maybe WAILA?
- Fluid output does not appear to work. Not sure why:
image

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/Crucible.java
```diff
@@ -0,0 +1,328 @@
...
+
+import java.util.*;
+
+import static io.github.pylonmc.pylon.base.util.BaseUtils.baseKey;
+
+public final class Crucible extends PylonBlock implements PylonMultiblock, PylonInteractBlock, PylonFluidTank, PylonCauldron, PylonBreakHandler, PylonTickingBlock {
+    public final int CAPACITY = getSettings().getOrThrow("capacity", ConfigAdapter.INT);
+    public final int SMELT_TIME = getSettings().getOrThrow("smelt-time", ConfigAdapter.INT);
+
+    private final Stack<ItemStack> contents = new Stack<>();
```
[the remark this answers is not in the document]
LordIdra: Just queuing one itemstack at a time feels totally fine to me. You can just up the amount of the itemstack if they try to add more of the same item. I feel like the bigger QoL problem is not being able to see what's inside the crucible, which becomes much more annoying to do if there is a queue of items. Also, cargo will be able to 'queue up' items to go in the crucible easily anyway.

Intybyte:
the method getTickInterval is called, and returns after how much time next tick is, normally it is saved on a constant variable inside the TickManager.

in this case, possibly in others, the getTickInterval output isn't constant, in this case it depends on the block below the crucible.

however the tickInterval used in the coroutine is set in stone, it can't be changed, even if it is supposed to be changed in order to tick faster, as maybe the player instead of using a torch below the crucible puts a lava bucket.

Proposal: every time after the tick is processed, instead of using a saved constant value, call getTickInterval every time

src/main/java/io/github/pylonmc/pylon/base/recipes/CrucibleRecipe.java
```diff
@@ -0,0 +1,134 @@
...
+            if (block == null) {
+                throw new UnsupportedOperationException("Can't make pylon item, no block or item available for such key");
+            }
+            ItemStack stack = new ItemStack(Material.BARRIER);
+            stack.setData(DataComponentTypes.ITEM_NAME, Component.text(block.getKey().toString()));
+            return stack;
+        }
+
+        return item.getItemStack();
+    }
```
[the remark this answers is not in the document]
LordIdra: (:+1: Addressed in convo below)

LordIdra (commented):
Just a couple more small things I noticed.

Another thing, the ticking rate seems to have gone up a LOT, like it chomps up cobblestone like crazy now. Assuming this isn't intended?

src/main/resources/recipes/minecraft/crafting_shaped.yml
```diff
@@ -336,6 +336,16 @@ pylonbase:mixing_pot:
...
+pylonbase:crucible:
+  pattern:
+    - "C C"
+    - "C C"
+    - "CCC"
+  key:
+    C: minecraft:brick
+  result: pylonbase:crucible
+  category: building
+
```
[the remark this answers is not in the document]
LordIdra: Not specifically. I was thinking it might be for skyblock, but I think there are a number of recipes you might need to modify anyway for that and you can easily do that now that recipes are data driven

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/Crucible.java
```diff
@@ -0,0 +1,290 @@
...
+                )),
+
+            PylonArgument.of("liquid_info", getFluidType() == null ?
+                Component.translatable("pylon.pylonbase.waila.crucible.liquid.empty") :
+                Component.translatable("pylon.pylonbase.waila.crucible.liquid.filled",
+                    PylonArgument.of("fluid", getFluidType().getName()),
+                    PylonArgument.of("amount", UnitFormat.MILLIBUCKETS.format(getFluidAmount()).decimalPlaces(1))
+                ))
+        ));
+    }
```
LordIdra: Ok just some quick WAILA stuff - I changed this a lot in my most recent PRs (which are now merged) so it would be good to update this for consistency. My suggestion:

State 1: Crucible | empty | empty
State 2: Crucible | 36x Cobblestone | empty
State 3: Crucible | empty | [||||||||||] 200/1000 mb (Lava)
State 4: Crucible | 36x Cobblestone | [||||||||||] 200/1000 mb (Lava)

You can create the fluid bar display using BaseUtils.createFluidAmountBar. If you're confused, check MixingPot on master - it works in a very similar way

src/main/resources/lang/en.yml
```diff
@@ -1477,6 +1486,13 @@ waila:
   fluid_tank:
     empty: "(empty)"
     filled: "| %bar%/%capacity% of %fluid%"
+  crucible:
+    item:
+      empty: "| Empty |"
+      stored: "| x%amount% %type% |"
+    liquid:
+      empty: "Empty"
```
Seggan: I would prefer "no fluid" or smth, it looks weird when no items and no fluid

src/main/resources/lang/en.yml
```diff
@@ -573,6 +573,15 @@ item:
       <arrow> <attr>Capacity:</attr> %capacity%
     waila: "Mixing Pot | %bar% (%fluid%)"
 
+  crucible:
+    name: "Crucible"
+    lore: |-
+      <arrow> Place items in the crucible using <insn>right click</insn>
+      <arrow> Can hold only one item type at a time
+      <arrow> The items will melt depending on the heat produced by the blow below the cauldron
+      <arrow> Requires a fire underneath to work
```
Seggan: Not just a fire

Issue #353, opened 2025-10-11
Add 'creative items' section which can only be seen by those with a specific permission


LordIdra:
For creative fluid source, etc

PR #451, opened 2025-11-15
Add Creative Items section in guide

Intybyte:
Closes https://github.com/pylonmc/pylon-base/issues/353
Requires https://github.com/pylonmc/pylon-core/pull/486


Intybyte:
> Should be a permission IMO

only reason it isn't a permission is that i have no idea what to name it

LordIdra:
Wait sorry I think that's only on my branch. Not merged yet but the permission can be used regardless.

Seggan (approved):

PR #456, opened 2025-11-20
Close lang tags

Intybyte:


Seggan (changes requested):
Do `<lang:xyz/>` instead

Seggan (approved):

PR #458, opened 2025-11-21
Vaan/fix/copper blocks axes

Intybyte:
Fixes https://github.com/pylonmc/pylon-core/issues/528


Intybyte:
> Agree with Seggan - maybe a PylonCopperWaxable or something, and then this logic can go in a listener

we already have a listener for that, why have 2 listener that do the same thing when it might create small problems, like which one gets called first, I agree on the idea of it being the default behaviour, but maybe we should just cancel the event as a default implementation of PylonInteractBlock

LordIdra:
> meant to say that we can use the PylonInteractBlock handler listener and cancel the event by default

This doesn't really feel intuitive to me tbh. If a pylon block is a hopper, I don't think that the default behaviour should be to cancel the interact event and force implementing blocks to then override and manually open the hopper interface.

I would just suggest solving it by creating an interface `PylonUnwaxable` or something (in core) that cancels all wax events using the logic you've added to `PylonCopperInteractBlock`. I would use a separate listener rather than overriding `PylonInteractBlock` (I think most other similar bits of logic do this). That shouldn't be too much work, it's just slightly reorganising what you've got already.

Intybyte:
> > meant to say that we can use the PylonInteractBlock handler listener and cancel the event by default
> 
> This doesn't really feel intuitive to me tbh. If a pylon block is a hopper, I don't think that the default behaviour should be to cancel the interact event and force implementing blocks to then override and manually open the hopper interface.
> 
> I would just suggest solving it by creating an interface `PylonUnwaxable` or something (in core) that cancels all wax events using the logic you've added to `PylonCopperInteractBlock`. I would use a separate listener rather than overriding `PylonInteractBlock` (I think most other similar bits of logic do this). That shouldn't be too much work, it's just slightly reorganising what you've got already.

Maybe not called PylonUnwaxable, as you can also remove rust if you wanted to and the interaction is the same, it seems useless to me having another event that does the exact same thing that interact does, and only creates confusion if both are implemented imo, one interface should interact with one event, it has always been like this, and I see no reason to change that now for this small edge case

PR #476, opened 2025-12-05
[ref:vaan/feature/procedural-crafting] Procedural Crafting

Intybyte:
Progress:

- [x] Recipe Type 
- [x] Make Item and Block
- [x] Item checks and processing logic
- [x] Display framework and progress

Stuff to decide:
- [ ] Lore and maybe change item name?
- [ ] Recipe
- [ ] Messages or particles or other

Example usage

```yml
pylonbase:iron_screw:
  ingredients:
    pattern:
      - "I  "
      - "   "
      - "   "
    key:
      I: pylonbase:iron_sheet
  results:
    - minecraft:stick: 8
  steps:
    - tool: minecraft:clock
      add_displays:
        - name: "base"
          material: light_gray_concrete
          position: [0, 0]
          scale: [1.0, 1.0]
    - tool: minecraft:iron_axe
      uses: 7
      remove_displays: ["base"]
      add_displays:
        - name: "screw1"
          material: light_gray_concrete
          position: [0.1, 0]
          scale: [0.08, 0.1]
          mirror_x: true
        - name: "screw2"
          material: light_gray_concrete
          position: [0.2, 0]
          scale: [0.08, 0.1]
          mirror_x: true
        - name: "screw3"
          material: light_gray_concrete
          position: [0.3, 0]
          scale: [0.08, 0.1]
          mirror_x: true
        - name: "screw4"
          material: light_gray_concrete
          position: [0.4, 0]
          scale: [0.08, 0.1]
          mirror_x: true
    - tool: pylonbase:iron_hammer
      uses: 8
```

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/recipes/intermediate/Step.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/recipes/intermediate/Step.java
@@ -165,8 +165,7 @@ public record Step(
     }
 
     public TranslatableComponent getUseTimes(int times) {
-        // todo: maybe generalize key to procedural-crafting as this will be reused for assembly table
-        return Component.translatable("pylon.pylonbase.guide.recipe.blueprint-workbench.used-times-lore")
+        return Component.translatable("pylon.pylonbase.guide.recipe.procedural-crafting.used-times-lore")
             .arguments(
                 PylonArgument.of(
                     "times",
--- a/src/main/resources/lang/en.yml
+++ b/src/main/resources/lang/en.yml
@@ -479,6 +479,11 @@ item:
     lore: |-
       <arrow> Smash stuff together hoping it works, jk, TODO
 
+  assembly_table:
+    name: "Blueprint Workbench"
+    lore: |-
+      <arrow> Smash stuff together hoping it works, jk, TODO
+
 
   grindstone:
     name: "Grindstone"
@@ -1571,7 +1576,7 @@ guide:
     casting: "<attr>Temperature:</attr> %temperature%"
     drilling: "<attr>Cycle time:</attr> %time%"
     molding: "<attr>Molding cycles:</attr> %molding-cycles%"
-    blueprint-workbench:
+    procedural-crafting:
       used-times-lore: "<gray> Use it %times% times"
 
 message:


Intybyte:
I feel like this is ready for review even if not fully complete, also from last testing it seems that on shutdown the entities aren't loaded back up properly, i have no idea what is causing it and it would be helpful if you guys had any idea

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/AssemblyTable.java
```diff
@@ -0,0 +1,405 @@
...
+                "# # # # # # # # #",
+                "# x x x # y y y #",
+                "# x x x $ y y y #",
+                "# x x x # y y y #",
+                "# # # # # # # # #"
+            )
+            .addIngredient('x', craftInventory)
+            .addIngredient('y', outputInventory)
+            .addIngredient('$', stepDisplay)
+            .addIngredient('#', GuiItems.backgroundBlack());
```
LordIdra: I would probably just use the normal background for this, it looks kinda weird with the texture pack:
image

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/AssemblyTable.java
```diff
@@ -0,0 +1,395 @@
...
+        this.currentRecipe = null;
+        this.currentProgress = null;
+        updateStep();
+    }
+
+    @SuppressWarnings("unused")
+    public AssemblyTable(@NotNull Block block, @NotNull PersistentDataContainer pdc) {
+        super(block);
+
+        byte[] craftingSerialized = pdc.get(CRAFTING_INVENTORY_KEY, PylonSerializers.BYTE_ARRAY);
```
LordIdra: We should have config adaptors for this rather than using byte arrays. (And if we don't we should create them)

LordIdra (changes requested):
Some general comments
- Instead of sending messages, we should give some visual indication. maybe particles of some kind? The only thing is they'd have to be per stage which is annoying
- Hammer sends 'cannot be used on this block' message - I assume this can be fixed later when PylonWorkbenchTool (or whatever its called) interface is introduced.
- Displays need to be yeeted when the craft is finished

I would suggest making all the recipes shapeless tbh. It feels more intuitive to me (there's no logical reason why the iron screws recipe should have to be in the top left at all - or even why there would ever be a fixed arrangement given it's a workbench where you're moving stuff around as you make it). It'll make automation 10x easier to implement as well. (It'll also make the code simpler)

I would suggest eventually replacing the workbench with a table display model + structure void. The working area feels awkwardly large, I think it only needs to be like 0.7x0.7

Instead of displaying the next step in the GUI, I would suggest using display entities instead so it's more immersive. Idea for this: above the model add the following (from top to bottom):
- Text display showing number of remaining uses
- Item display with the item that needs to be used next
- Small text display with the name of the item that needs to be used next
(All should be billboards so they face the player all the time)

I think this worbench should only have an input section in the UI, and the output should just be dropped in world. Then, we could introduce some kind of 'Blueprint Workbench' which holds a specific workbench recipe and has both an input/output. This would make full automation feasible. And also it would be consistent with other early-game machines as all of them drop stuff in-world

Btw you're probably going to need some kind of Z-index for displays because some might end up overlapping each other

have only very lightly skimmed the code since I imagine a lot will still change

Intybyte:
>     * Instead of sending messages, we should give some visual indication. maybe particles of some kind? The only thing is they'd have to be per stage which is annoying

It is going to be annoying regardless, i don't want it to mess with action bars because most plugins already use it, nor waila, so message is the best thing, i don't want to mute my pc every time i am crafting a recipe with tons of steps either, message is the best thing imo, or maybe a low volume sound if you prefer, or we can have both, one of my ideas was adding an optional sound per step after all

>     * Hammer sends 'cannot be used on this block' message - I assume this can be fixed later when PylonWorkbenchTool (or whatever its called) interface is introduced.

Yeah i have no idea how we are going to do that though, what was your idea for PylonWorkbenchTool

>     * Displays need to be yeeted when the craft is finished

It was a bug, has been fixed

> I would suggest eventually replacing the workbench with a table display model + structure void. The working area feels awkwardly large, I think it only needs to be like 0.7x0.7

Is this the same for the blueprint table? if you want you can commit how you want it to look like directly tbh
 
> Instead of displaying the next step in the GUI, I would suggest using display entities instead so it's more immersive. Idea for this: above the model add the following (from top to bottom):

Added, it always faces north regardless which is a bit annoying, the display in the GUI is staying because tbh, it is convenient to not spin around the table sometimes, also it causes no harm keeping it 



Once this is done i can start work on the assembly table

LordIdra:
> It is going to be annoying regardless, i don't want it to mess with action bars because most plugins already use it, nor waila, so message is the best thing, i don't want to mute my pc every time i am crafting a recipe with tons of steps either, message is the best thing imo, or maybe a low volume sound if you prefer, or we can have both, one of my ideas was adding an optional sound per step after all

Why not particles as suggested? Message just seems really jank and will be extremely annoying in a multiplayer environment where you also want to see what other players are typing.

> Yeah i have no idea how we are going to do that though, what was your idea for PylonWorkbenchTool

See the last message in the procedural crafting thread

> Is this the same for the blueprint table? if you want you can commit how you want it to look like directly tbh

I think yes blueprint table should also have a smaller area as it feels weird to have it this big imo. If you don't want to do it then I can do it in another pr, but tbh it's not really hard just fiddly and I don't have a clear picture for how it should look either.

> Added, it always faces north regardless which is a bit annoying, the display in the GUI is staying because tbh, it is convenient to not spin around the table sometimes, also it causes no harm keeping it 

This is why I said '(All should be billboards so they face the player all the time)'. Use the billboard setting.

Intybyte:
Yeah let's make the models in other PRs, didn't know what billboard was, will look for the other stuff later

src/main/java/io/github/pylonmc/pylon/base/recipes/intermediate/RecipeFormation.java
```diff
@@ -0,0 +1,72 @@
...
+            if (choice == null || !choice.test(stack)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    public RecipeChoice getChoice(int x, int y) {
+        var character = this.rows[y].charAt(x);
```
Seggan: dude was too lazy to type one extra letter

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/BlueprintWorkbench.java
```diff
@@ -0,0 +1,274 @@
...
+    private void updateStepItem() {
+        ItemStack stack;
+        if (currentRecipe == null || currentProgress == null) {
+            stack = GuiItems.background().getItemProvider().get();
+        } else {
+            Step current = this.getStep();
+            stack = current.asStack(current.uses() - currentProgress.getUsedAmount());
+        }
+
+        this.stepDisplay.setItem(UpdateReason.SUPPRESSED, 0, stack);
```
LordIdra: Why is stepDisplay a virtual inventory? (Actually - maybe you don't know about invui's items? Have a read of the invui docs - https://docs.xenondevs.xyz/invui/ - using a custom item is a much nicer approach than having a virtual inventory)

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/ProceduralCraftingTable.java
```diff
@@ -0,0 +1,173 @@
...
+import xyz.xenondevs.invui.inventory.event.UpdateReason;
+
+import java.util.ArrayList;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+import static io.github.pylonmc.pylon.base.util.BaseUtils.baseKey;
+
+@Getter
+public abstract class ProceduralCraftingTable<T extends Step.StepsHolder & PylonRecipe> extends PylonBlock implements PylonEntityHolderBlock {
```
LordIdra: This class seems rather overengineered, probably partially owing to the generic, which can be removed now. I would suggest just collapsing this into BlueprintWorkbench (assuming you're not adding AssemblyTable in this PR). If there is a good abstraction we can make it later.

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/ProceduralCraftingTable.java
```diff
@@ -0,0 +1,173 @@
...
+    protected final VirtualInventory craftInventory;
+    protected T currentRecipe;
+    protected Step.ActionStep currentProgress;
+
+    protected final Step.StateDisplay currentStateDisplay;
+
+    public ProceduralCraftingTable(@NotNull Block block, @NotNull BlockCreateContext context) {
+        super(block, context);
+
+        this.craftInventory = new VirtualInventory(9);
```
LordIdra: Bit of a nitpick but prefer craftingInventory here, as craftInventory implies NMS

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/ProceduralCraftingTable.java
```diff
@@ -0,0 +1,173 @@
...
+        this.currentStateDisplay = Step.StateDisplay.init(this);
+    }
+
+    public ProceduralCraftingTable(@NotNull Block block, @NotNull PersistentDataContainer pdc) {
+        super(block, pdc);
+
+        this.craftInventory = pdc.get(CRAFTING_INVENTORY_KEY, PylonSerializers.VIRTUAL_INVENTORY);
+
+        NamespacedKey currentRecipeKey = pdc.get(CURRENT_RECIPE_KEY, PylonSerializers.NAMESPACED_KEY);
+        this.currentRecipe = currentRecipeKey == null ? null : deserializeRecipeKey(currentRecipeKey);
```
LordIdra: You can use a KEYED.keyedTypeFrom serializer here - example from RecipeProcessorPersistentDataType:
```
val recipePDT = PylonSerializers.KEYED.keyedTypeFrom { recipeType.getRecipeOrThrow(it) }
```
It'll simplify a lot of the serde logic which is nice

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/ProceduralCraftingTable.java
```diff
@@ -0,0 +1,173 @@
...
+            case NEXT_STEP, NEXT_PROGRESS -> updateStep();
+            case COMPLETED_RECIPE -> completeRecipe();
+        }
+
+        return result.isSuccess();
+    }
+
+    protected static void clearInventory(VirtualInventory inventory) {
+        for (int i = 0; i < 9; i++) {
+            inventory.setItem(UpdateReason.SUPPRESSED, i, null);
```
LordIdra: For `inventory.setItem` calls you can now use `new MachineUpdateReason()` which I added recently. (Won't comment on the other calls)

src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/ProceduralCraftingTable.java
```diff
@@ -0,0 +1,173 @@
...
+            if (data.mirrorZ()) {
+                addEntityIfMissing(name + "$mirror_z", () -> new BlockDisplayBuilder()
+                    .material(data.material())
+                    .transformation(new TransformBuilder()
+                        .translate(positions[0], 0, -positions[1])
+                        .scale(scale[0], 0, scale[1])
+                    )
+                    .build(up)
+                );
+            }
```
LordIdra: I don't think this will display the correct behaviour if both X and Z are mirrored - in that case there should be 4 entities.

src/main/java/io/github/pylonmc/pylon/base/recipes/intermediate/DisplayData.java
```diff
@@ -0,0 +1,61 @@
...
+import io.github.pylonmc.pylon.base.recipes.BlueprintWorkbenchRecipe;
+import io.github.pylonmc.pylon.core.config.adapter.ConfigAdapter;
+import io.github.pylonmc.pylon.core.config.adapter.MapConfigAdapter;
+import org.bukkit.Material;
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.reflect.Type;
+import java.util.List;
+
+public record DisplayData(
```
LordIdra: Should probably be called something a bit less generic, eg AssemblyTableDisplayData (DisplayData could mean almost anything)

PR #482, opened 2025-12-11
[ref:vaan/api/falling-py-block] Gravity aware bronze anvil

Intybyte:
Depends on: https://github.com/pylonmc/pylon-core/pull/520

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/BronzeAnvil.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/BronzeAnvil.java
@@ -21,7 +21,6 @@ import io.github.pylonmc.pylon.core.entity.display.ItemDisplayBuilder;
 import io.github.pylonmc.pylon.core.entity.display.transform.TransformBuilder;
 import io.github.pylonmc.pylon.core.item.PylonItem;
 import io.github.pylonmc.pylon.core.util.PylonUtils;
-import io.github.pylonmc.pylon.core.util.position.BlockPosition;
 import net.kyori.adventure.sound.Sound;
 import org.bukkit.Location;
 import org.bukkit.NamespacedKey;
@@ -39,9 +38,7 @@ import org.bukkit.persistence.PersistentDataContainer;
 import org.jetbrains.annotations.NotNull;
 import org.joml.Matrix4f;
 
-import java.util.HashMap;
 import java.util.List;
-import java.util.UUID;
 import java.util.concurrent.ThreadLocalRandom;
 
 import static io.github.pylonmc.pylon.base.util.BaseUtils.baseKey;


src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/BronzeAnvil.java
```diff
@@ -182,13 +192,42 @@ public void tick(double deltaSeconds) {
...
+            .build(getBlock().getLocation().toCenterLocation());
+
+        itemDisplay.getPersistentDataContainer().set(PylonEntityHolderBlock.getBlockKey(), PylonSerializers.BLOCK_POSITION, new BlockPosition(event.getBlock()));
+        map.put("item", itemDisplay.getUniqueId());
+
+        entity.getBlockData().set(
+            PylonEntityHolderBlock.getEntityKey(),
+            PylonEntityHolderBlock.getEntityType(),
+            map
+        );
```
LordIdra: How come you can't just use `addEntity` here? I know it's not in BlockStorage yet but I think it should still work.

LordIdra (changes requested):

src/main/java/io/github/pylonmc/pylon/base/content/machines/smelting/BronzeAnvil.java
Intybyte: the block doesn't exist at that time yet, as written in the docs, also everything is happening before deserialization, if i were to add it with addEntity it would get lost later

LordIdra: How can it not exist when this method is being called on that very block? Oh or is it the old block? In that case would it maybe sense to change this to be called after the block is created instead? That seems like it would be cleaner

Intybyte: > How can it not exist when this method is being called on that very block? Oh or is it the old block? In that case would it maybe sense to change this to be called after the block is created instead? That seems like it would be cleaner

I use the constructor to make it on the fly, it isn't actually anywhere, in fact you need to threat the onFallStop method as a static one, all the other informations aren't avaliable, reason why if you see, I need to put in the entity pdc stuff like the direction and such.

I can call it after if you prefer, however this has the disadvantage that you cannot cancel the event anymore in case you needed to do something more complex, and probably some more internals in BlockStorage will need to be exposed

Intybyte: fixed and tested

Intybyte:
> Bronze anvil errors about 10 seconds after falling with:
> 
> ```
> [22:11:07 ERROR]: [PylonCore] Error when handling block(pylonbase:bronze_anvil, Location{world=CraftWorld{name=world},x=-382.0,y=75.0,z=-682.0,pitch=0.0,yaw=0.0}) ticking: Entity item does not exist or is not of type ItemDisplay
> [22:11:07 WARN]: java.lang.IllegalArgumentException: Entity item does not exist or is not of type ItemDisplay
> [22:11:07 WARN]:        at pylon-core-1.0.0-SNAPSHOT.jar//io.github.pylonmc.pylon.core.block.base.PylonEntityHolderBlock.getHeldEntityOrThrow(PylonEntityHolderBlock.kt:75)
> [22:11:07 WARN]:        at pylon-base-1.0.0-SNAPSHOT.jar//io.github.pylonmc.pylon.base.content.machines.smelting.BronzeAnvil.getItemDisplay(BronzeAnvil.java:218)
> [22:11:07 WARN]:        at pylon-base-1.0.0-SNAPSHOT.jar//io.github.pylonmc.pylon.base.content.machines.smelting.BronzeAnvil.tick(BronzeAnvil.java:180)
> [22:11:07 WARN]:        at pylon-core-1.0.0-SNAPSHOT.jar//io.github.pylonmc.pylon.core.block.TickManager$startTicker$1.invokeSuspend(TickManager.kt:75)
> [22:11:07 WARN]:        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
> [22:11:07 WARN]:        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
> [22:11:07 WARN]:        at org.bukkit.craftbukkit.scheduler.CraftTask.run(CraftTask.java:78)
> [22:11:07 WARN]:        at org.bukkit.craftbukkit.scheduler.CraftScheduler.mainThreadHeartbeat(CraftScheduler.java:474)
> [22:11:07 WARN]:        at net.minecraft.server.MinecraftServer.tickChildren(MinecraftServer.java:1744)
> [22:11:07 WARN]:        at net.minecraft.server.MinecraftServer.tickServer(MinecraftServer.java:1618)
> [22:11:07 WARN]:        at net.minecraft.server.dedicated.DedicatedServer.tickServer(DedicatedServer.java:430)
> [22:11:07 WARN]:        at net.minecraft.server.MinecraftServer.runServer(MinecraftServer.java:1338)
> [22:11:07 WARN]:        at net.minecraft.server.MinecraftServer.lambda$spin$2(MinecraftServer.java:384)
> [22:11:07 WARN]:        at java.base/java.lang.Thread.run(Thread.java:1583)
> ```
> 
> Also, the anvil drops an item as if it was broken when it stops falling

Fixed, the problem was that onDeserialize of tickable blocks it ticks when the entity doesn't exist yet, because now the deserialization is called before the actual method, and at that time the entity wasn't created yet

LordIdra (approved):

Seggan (approved):

PR #497, opened 2025-12-21
[ref:idras-little-fixes] Idras little fixes

LordIdra:
Just a ton of small things both codewise and in-game, there were too many to really split it up into many PRs without completely losing my sanity (sorry). Here is the list of changes. Half of this typed out on 1 hour sleep sorry if no makey sense
- Fixes yet more references to enriched netherrack that were not changed when it was renamed
- Move fields to the start of each block class for consistency
- Store item display stacks as `ItemStackBuilder` to remove boilerplate
- Updates blocks to use new blockface methods added in core PR
- Updates blocks to use new PylonDirectionalBlock design
- Adds PylonDirectionBlock to a lot of blocks which should have it but don't
- Makes diesel machines fully respect their placed direction
- Adds diesel / hydraulic fluid buffer settings and placeholders to all diesel / hydraulic machines
- Adds fluid display bars to all hydraulic machine WAILAs
- Adds displays to creative fluid source which change material to indicate the fluid
- Adds fluid type to creative fluid source WAILA
- Removes deltaSeconds usage in line with companion core PR
- Makes settings not static where they don't need to be static
- Changes some fluid blocks (eg filter, valve) to place in a more consistent orientation (input facing towards player)
- Changes hydraulic / diesel machines to consume hydraulic fluid / diesel gradually, instead of all in one go (where practical)
- Pretty much rewrites strainer to take advantage of all the new stuff
- Changes fluid blocks to make use of fluidTickInterval as in companion core PR
- Makes coal-fired purification tower use a registry instead of reading from settings
- Adds WAILA override to hydraulic core drill hatch casings
- Made hydraulic grindstone turner not a multiblock
- Make hydraulic mixing attachment not a multiblock
- Fix hydraulic hammer head going down too far
- Made hydraulic press piston not a multiblock
- Removed pylon craft events as per companion core PR
- Make mixing pot not a multiblock
- Reworked hammer logic to be a bit less cursed and not consume and then drop the items it doesn't use (this looked really awkward with the hammer press)
- Lots of other small code style / formatting changes to make everything consistent

Tested pretty much all blocks (multiple times) but have likely introduced plenty of bugs due to the scope of the changes :(

Pretty much ignored smeltery here, still can't conjure the courage to go and fight that particular boss battle

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/base/content/building/Pedestal.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/building/Pedestal.java
@@ -9,6 +9,8 @@ import io.github.pylonmc.pylon.core.block.context.BlockCreateContext;
 import io.github.pylonmc.pylon.core.datatypes.PylonSerializers;
 import io.github.pylonmc.pylon.core.entity.display.ItemDisplayBuilder;
 import io.github.pylonmc.pylon.core.entity.display.transform.TransformBuilder;
+import lombok.Getter;
+import lombok.Setter;
 import org.bukkit.NamespacedKey;
 import org.bukkit.block.Block;
 import org.bukkit.entity.ItemDisplay;
@@ -29,7 +31,8 @@ public class Pedestal extends PylonBlock implements PylonEntityHolderBlock, Pylo
     private static final NamespacedKey ROTATION_KEY = baseKey("rotation");
     private static final NamespacedKey LOCKED_KEY = baseKey("locked");
     private double rotation;
-    public boolean locked;
+    @Getter @Setter
+    private boolean locked;
 
     @SuppressWarnings("unused")
     public Pedestal(@NotNull Block block, @NotNull BlockCreateContext context) {
--- a/src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/MagicAltar.java
+++ b/src/main/java/io/github/pylonmc/pylon/base/content/machines/simple/MagicAltar.java
@@ -119,7 +119,7 @@ public class MagicAltar extends PylonBlock
             itemDisplay.setItemStack(catalyst.asQuantity(1));
             catalyst.subtract();
             for (Pedestal pedestal : getPedestals()) {
-                pedestal.locked = true;
+                pedestal.setLocked(true);
             }
             startRecipe(recipe, recipe.timeSeconds() * 20);
             break;
@@ -195,7 +195,7 @@ public class MagicAltar extends PylonBlock
     public void onRecipeFinished(@NotNull MagicAltarRecipe recipe) {
         for (Pedestal pedestal : getPedestals()) {
             pedestal.getItemDisplay().setItemStack(null);
-            pedestal.locked = false;
+            pedestal.setLocked(false);
         }
 
         getItemDisplay().setItemStack(recipe.result());
@@ -212,7 +212,7 @@ public class MagicAltar extends PylonBlock
 
         for (Pedestal pedestal : getPedestals()) {
             if (pedestal != null) {
-                pedestal.locked = false;
+                pedestal.setLocked(false);
             }
         }
 


src/main/java/io/github/pylonmc/pylon/base/content/machines/hydraulics/HydraulicCoreDrillHatch.java
```diff
@@ -47,6 +49,12 @@ public boolean isPartOfMultiblock(@NotNull Block otherBlock) {
         return getBlock().getRelative(BlockFace.UP) == otherBlock;
     }
 
+    @Override
+    public void onMultiblockRefreshed() {
+        Waila.addWailaOverride(new BlockPosition(getBlock().getRelative(BlockFace.UP)), this::getWaila);
```
Seggan: When this block is broken the override won't disappear

src/main/java/io/github/pylonmc/pylon/base/content/building/Pedestal.java
```diff
@@ -30,8 +29,7 @@ public class Pedestal extends PylonBlock implements PylonEntityHolderBlock, Pylo
     private static final NamespacedKey ROTATION_KEY = baseKey("rotation");
     private static final NamespacedKey LOCKED_KEY = baseKey("locked");
     private double rotation;
-    @Setter
-    private boolean locked;
+    public boolean locked;
```
Seggan: Why has this been made public?

src/main/java/io/github/pylonmc/pylon/base/content/tools/HydraulicCannon.java
```diff
@@ -53,6 +54,12 @@ public HydraulicCannon(@NotNull ItemStack stack) {
 
     @Override
     public @NotNull List<@NotNull PylonArgument> getPlaceholders() {
+        Bukkit.getLogger().severe("a " + getHydraulicFluid() + " " + HYDRAULIC_FLUID_CAPACITY + " " + BaseUtils.createFluidAmountBar(
```
Seggan: Forgot to remove debug line?

Seggan (changes requested):

Seggan:
I've taken the liberty to de-delta the smeltery in my own PR

src/main/java/io/github/pylonmc/pylon/base/content/machines/hydraulics/HydraulicCoreDrillHatch.java
LordIdra: I have removed this block entirely in a later PR so will leave this be

LordIdra:
> I've taken the liberty to de-delta the smeltery in my own PR

Do you want me to remove the changes I made here then?

Seggan:
Yeah I've changed it more comprehensively

src/main/java/io/github/pylonmc/pylon/base/content/building/Pedestal.java
Seggan: why not the getter/setter then

LordIdra: why not public? We use public basically everywhere so I just made it public. I don't think it matters whether it's public or has @getter @setter especially in this context

LordIdra:
Ok I'm gonna have to keep the changes since I've removed deltaSeconds entirely in core, so this won't compile in that case. Is it OK if we keep them, and then you merge this PR (I'm assuming this PR will be merged into master before yours) into your PR, or just override my changes completely?

Seggan:
Yeyeye ofc

src/main/java/io/github/pylonmc/pylon/base/content/building/Pedestal.java
Seggan: 1. Consistency
2. Best practices
3. What else do we use lombok for?

LordIdra: Are we really going to argue about this? Does it really matter? Does it affect anything? Most fields across base are just public in case someone wants to access something for some unknown reason.

PR #502, opened 2025-12-25
[ref:more-cargo-blocks] Breakers and miners

LordIdra:
- Remove hydraulic excavator
- Add hydraulic breaker
- Add hydraulic miner
- Add diesel breaker
- Add diesel miner


jiajia06403:
miner break bedrock (very fast! with any tool!)

LordIdra:
> breaker breaks block without outputing item

Lmao what the hell, how did I miss that

All issues should be fixed, thanks for testing :+1:

Seggan (commented):
LGTM otherwise

PR #504, opened 2025-12-27
[ref:breakers-and-miners] Biodiesel manufacturing

LordIdra:
- Removes hydraulic core drill hatches and replaces them with generic fluid input/output hatches
- Adds item input/output hatches
- Adds fermenter
- Adds biorefinery


balugaq:
Bug1: the fermenter waila should display 256 instead of 257
QQ_1767365332196

PR #515, opened 2026-01-01
ADD MORE CARGO THINGS

LordIdra:
- Add cargo monitor
- Add cargo gate
- Add cargo meter
- Add cargo accumulator
- Add cargo fluid accumulator
- Add directional block data to solar lens for texture packs
- Overhaul fluid meter (yes, in an 'even more cargo blocks' PR - I wanted to make it consistent with the cargo meter)
- Change 'fluid machines' category icon to light blue stained glass instead of a cauldron (none of the items in fluid machines are a cauldron lol)
- Change visuals of most cargo blocks to be more coherent
- Change visuals of fluid meter, filter, and valve to be more coherent (and consistent with cargo)
- Change fluid filter to swap materials to indicate status instead of brightness (it looked really weird at night)
- Fix stupid issue where cargo extractors/inserters would connect to the face *behind* them instead of *in front* of them after a cargo block was connected then disconnected
- Fix being able to TRANSMUTE fluid contained in a fluid tank by removing one casing, then replacing it and piping in a new fluid
- Fix hydraulic farmer not breaking sugarcane
- Fix press recipe type not being set
- Fix press freezing after loading if a recipe was in progress at the time of unload


Seggan (commented):
Turing incompleteness aside, the cargo monitor doesn't show any item if it passes through it immediately, giving the false impression there's no item in the network.

Code LGTM.

PR #524, opened 2026-01-06
[ref:chore/1.21.11] Logic gates go brr

Seggan:
Added cargo overflow gate, and the cargo versions of the infinite fluid source and voider


Seggan:
> Yeah, but the splitter does arbitrary splitting. Maybe it's just me but it feels better to me to have one purpose for each machine - this is a sort of weird middle-ground between splitting and overflow where it can only split in 1:1 but also has an overflow mechanic added on top, it feels a little bit arbitrary

I may want to supply 2 machines equally but allow the backing up of one (or both), especially earlygame where I may not run machines constantly. This is literally just the factorio splitter lol (no I haven't been playing too much factorio)

PR #561, opened 2026-01-20
Item/research reorganisation

LordIdra:
- Groups/orders items more logically in the guide
- Reorders fluids in the guide to make more sense
- Creates several new subcategories to better split up items
- Renames farmer talisman -> farming talisman
- Renames magic altar / magic pedestal -> shimmer altar / shimmer pedestal ('magic altar' is a bit too generic and we might want to add more advanced magic ritual stuff in future)
- Renames item magnet -> shimmer magnet (to clearly show it's magic and not related to cobalt/nickel)
- Changes researches to use Pylon items as icons
- Revamps researches so they make more sense with all the new items and show progression more clearly
- A few small recipe changes


src/main/resources/recipes/pylonbase/magic_altar.yml
Seggan: Fogor rename file

PR #616, opened 2026-02-22
Add check for world borders

### LINKED ISSUE #615: Hydraulic Miner breaks blocks outside the world border
Image

balugaq:
Closes #615


Seggan (changes requested):
We have a method in RebarUtils already that *should* handle all that + enchantments and stuff wrt item damage

Seggan (approved):

PR #618, opened 2026-02-22
Fix pit kiln not breaking charcoal blocks

### LINKED ISSUE #614: Pit Kiln does not destory charcoal correctly

OhmV-IR:
Charcoal blocks are actually a pylon block, so when we tried to break it by calling the BlockBreakBlockEvent and settings its type to air, rebar cancels the event. Fixed by checking if each block is a rebar block and if it is break via blockstorage, otherwise use vanilla method.
Closes #614


JustAHuman-xD:
Not how this should be fixed, we need to handle BlockBreakByBlockEvent in rebar, we can change which event we fire for rebar blocks for this but we do need to handle that block in rebar

Seggan (changes requested):
What Justin said

PR #670, opened 2026-02-23
fix(anvil): prevent vanilla anvil GUI from opening on bronze anvil

### LINKED ISSUE #649: Steel working bug
Jackk
OP
 — 2:14 PM
Using the bronze anvil for the first time, not sure if im doing it right. But even using the tongs to grab the hot thing still hurts.

But mainly having the bronze anvil be a real anvil is kinda yek, since you have to place the hot steel on it and then it opens the anvil. If it was a block that looked like an anvil or something with no inventory it would be better.
Seggan

 — 2:19 PM
It opens the anvil? Hmm that should not happen
Jackk
OP
 — 2:29 PM
ya
Seggan

 — 2:30 PM
Yeah that's a bug
As for the tongs, you need it in offhand

SamuelJoseph23:
Added setUseInteractedBlock(DENY) to BronzeAnvil.onInteract() at NORMAL priority. Without this, right-clicking the bronze anvil opens the vanilla Minecraft anvil GUI instead of only using the custom item placement logic. Fixes #649

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/content/machines/smelting/BronzeAnvil.java
+++ b/src/main/java/io/github/pylonmc/pylon/content/machines/smelting/BronzeAnvil.java
@@ -8,6 +8,7 @@ import io.github.pylonmc.pylon.content.tools.Hammer;
 import io.github.pylonmc.pylon.util.PylonUtils;
 import io.github.pylonmc.rebar.block.RebarBlock;
 import io.github.pylonmc.rebar.block.base.*;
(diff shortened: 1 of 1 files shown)


JustAHuman-xD:
The solution to this should just be implementing NoVanillaContainerBlock afaik

SamuelJoseph23:
@JustAHuman-xD Good call! Since the Bukkit Anvil is technically a Container, I've implemented RebarNoVanillaContainerBlock on the BronzeAnvil class. This cleanly suppresses the InventoryOpenEvent without breaking the block interaction mechanics.

PR #677, opened 2026-02-23
fix(loupe): suppress examining message when on cooldown

### LINKED ISSUE #611: Loupe sends examining message even if on cooldown

SamuelJoseph23:
Added player.hasCooldown(getStack()) check to onUsedToClick() early return. The USE_COOLDOWN data component only prevents the consume event, not the initial interact event, so the examining action bar message and sound were firing even on cooldown. Fixes #611


LordIdra:
I can resolve conflicts. Happy to merge. The setCooldown could be extracted to after the ifelse block but I'm not bothered enough about that to block

PR #685, opened 2026-02-23
Fix core drills running after multiblock break.

### LINKED ISSUE #628: Manual drills keep working even if destroyed
It even spits out an item, which lets you easily run one drill at 3-4x the speed at the expense of pickaxe durability

Image

OhmV-IR:
Closes #628


Seggan (changes requested):
Still produces particles and an item

Image

OhmV-IR:
I wasn't able to reproduce that issue. All of the particles stopped for me and there was no item dropped or particles made. Did find one where the lever got stuck down tho so I fixed that. https://youtu.be/VUVCu6Fhog4

PR #690, opened 2026-02-23
Vaan/feature/feature collection

### LINKED ISSUE #39: Tape measure
Probably using display entities

### LINKED ISSUE #653: Upgraded loupe with less cooldown
Idra

OP
 — Yesterday at 6:17 PM
because it's annoying

### LINKED ISSUE #665: Glass Breaker
Nycto

OP
 — 2026-02-09 5:33 PM
For the love of god any tool that can instant break glass/beacons and the few “fragile” blocks that take forever to look. Bonus points if its a tome or enchant applicable to any item
JustAHuman

 — 2026-02-09 5:37 PM
iirc @Vaan1310 (Intybyte) you were already doing this right 
or was that sir seggan
Seggan

 — 2026-02-09 5:39 PM
uhh idk
JustAHuman

 — 2026-02-09 5:40 PM
I swear someone brought it up....
[⁠🔒⎪dev-only⁠
](https://discord.com/channels/1329177304857055273/1329177716582256808/1460317949750546502)ahhhh
Nycto

OP
 — 2026-02-09 5:41 PM
And the Chaos begins….
Vaan1310 (Intybyte)

 — 2026-02-09 5:49 PM
one of the times i asked for stuff and got ignored lol (or more realistically everyone was busy)

### LINKED ISSUE #683: Bed Roll
AriaNight

OP
 — 2026-02-21 11:15 AM
A bed you can take with you and when you place it to sleep it skips the night and goes back into ur inv. Also doesn't set a respawn point. Only the bed does.

To loop players: What if you could loop staff players to get special items!
Ohm

 — Yesterday at 10:03 PM
Rebar doesn't provide a first-party rank plugin so we have no way to know who is staff and who isn't. Also many servers use their own plugins for moderation so I think that kills that part. The bed roll doesn't seem more useful than the vanilla bed so I don't think we'll add that either
Idra

 — Yesterday at 10:49 PM
I think sleeping bag would be neat tbh
I'm not against another thing for casual players
Seggan

 — Yesterday at 10:52 PM
same
and we can detect opped players
if you manage to catch an opped player it could be cool to loupe them
JustAHuman

 — 2:29 AM
bed roll is definitely more useful than a vanilla bed, at least to some people/situations 
The comforts mod has 200+ million downloads for a reason

Intybyte:
Fixes #665 
Fixes #653


JustAHuman-xD (changes requested):

src/main/java/io/github/pylonmc/pylon/content/tools/TapeMeasure.java
LordIdra: Just doing a skim and I think it would be nice if:
1) There was a display entity line drawn from the origin to where you're currently looking
2) The tape measure could handle sub-block distances
3) The distance was shown live in the actionbar/bossbar. This might conflict with WAILA so not sure but maybe worth looking into

If you want to do that here that would be great, otherwise we can open issues

src/main/resources/lang/en.yml
```diff
@@ -564,6 +582,23 @@ item:
...
+      <arrow> An upgraded version of the loupe for faster scanning
+      <arrow> Examining items/blocks with the loupe yields research points
+      <arrow> The rarer the item/block, the more times it can be examined
+      <arrow> <red>The examined item/block will be consumed</red>
+      <arrow> <insn>Hold right click</insn> on a block or dropped item to examine it
+
+  tape_measure:
+    name: "Tape Measure"
+    lore: |-
+      <arrow> Measures the distance between two blocks in the same world
```
LordIdra: 'in the same world' probably redundant

src/main/resources/lang/en.yml
```diff
@@ -152,6 +152,24 @@ item:
...
+  glass_breaker:
+    name: "Glass Breaker"
+    lore: |-
+      <arrow> Breaks glass like material faster, includes:
+      <arrow> - All types of glass blocks
+      <arrow> - All types of glass panes
+      <arrow> - Sea Lantern
+      <arrow> - Glowstone Block
+      <arrow> - Redstone Lamp
+      <arrow> - Beacon
```
LordIdra, line 166: New lore suggestion:
```
<arrow> Breaks fragile blocks faster
<arrow> Works on <yellow>Glass Blocks</yellow>, <yellow>Glass Panes</yellow>, <yellow>Sea Lanterns</yellow>, <yellow>Glowstone Blocks</yellow>, <yellow>Redstone Lamps</yellow>, and <yellow>Beacons</yellow>
```
(2nd line looks way longer than it is because the tags)

LordIdra (changes requested):
( Bed roll )
I think sleeping bag would be a better name for bed roll tbh

Clicking 'leave bed' while using bed roll does not remove the bed


( Tape measure )

I was thinking with the tape measure it would draw the line as you dragged it - so every tick the line would update to end where you are looking. This would allow you to remove the 'position 1 set' and 'position 2 set' messages too

No way to clear tape measure line - maybe do when the player scrolls off it?

Would be better I think to calculate the distance immediately when position 2 is set instead of waiting for shift+right click

What would be really nice here is to use a text display which constantly teleports to where the player is looking and displays the current distance. This would make the tape measure a lot more useful

May be just me but left click for pos1 and right click for pos2 feels more natural?


Apologies I missed this for so long...

balugaq (approved):

Seggan (approved):

LordIdra:
(Just to flag - working on adding tiers for the loupe / glass breaker just haven't pushed yet, so don't merge yet)

PR #707, opened 2026-02-25
Add grindstone waila

balugaq:


LordIdra (changes requested):
This doesn't solve the issue of not being able to see what is inside a grindstone. I would suggest instead:

`Grindstone | Cobblestone x44`
and
`Grindstone | Cobblestone x44 | [||||||||||] 50%`

PR #761, opened 2026-02-27
Fix breakers sometimes no longer working

balugaq:


LordIdra (approved):
Ah that was a silly oversight

PR #784, opened 2026-03-01
Fix base ci

OhmV-IR:


OhmV-IR:
Tested by https://github.com/pylonmc/pylon/actions/runs/22535574760

LordIdra:
what was the problem and how does this solve it?

PR #863, opened 2026-03-08
Polish smeltery

### LINKED ISSUE #580: Interpolate smeltery display color and position

### LINKED ISSUE #783: Smeltery Damage
**Issue reported by: nyctottv:**
Make the smeltery do damage to the player when in it, you could also add blood like Tinkers Construct has but thats optional

Seggan:
- Updated smeltery recipes
- Removed Smeltery Caster and made it a separate block called the Casting Unit that requires the fluid piped in
- Smeltery Input and Output Hatches now require the correct temperature
- Added casts, for casting liquid metals to nuggets, ingots, blocks, sheets, and pipes (pipes use the crafting ratio, so pipe bender is still more efficient)


src/main/java/io/github/pylonmc/pylon/content/machines/smelting/CastingUnit.java
```diff
@@ -0,0 +1,232 @@
...
+    private static final ItemStack CAST_BORDER = ItemStackBuilder.gui(Material.GREEN_STAINED_GLASS_PANE, pylonKey("cast"))
+            .name(Component.translatable("pylon.gui.cast-border"))
+            .build();
+
+    @Override
+    public @NotNull Gui createGui() {
+        return Gui.builder()
+                .setStructure(
+                        "# # # # # c c c #",
+                        "# # # i # c x c #",
```
Intybyte: at this point surround `i` with `GuiItems#input`

build.gradle.kts
```diff
@@ -30,6 +30,9 @@ repositories {
     maven("https://maven.pvphub.me/tofaa") {
         name = "EntityLib"
     }
+    maven("https://repo.codemc.io/repository/maven-releases/") {
+        name = "CodeMC"
+    }
```
Intybyte: why was this added?

Intybyte (changes requested):
Reviewed, works ingame, also are we sure we should drop the documenation of said recipes? I mean they are easy to understand even without it

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/CastingUnit.java
```diff
@@ -0,0 +1,232 @@
...
+
+            notifyWindows();
+        }
+    }
+
+    private final CastingControlItem castingControlItem = new CastingControlItem();
+
+    private static final ItemStack CAST_BORDER = ItemStackBuilder.gui(Material.GREEN_STAINED_GLASS_PANE, pylonKey("cast"))
+            .name(Component.translatable("pylon.gui.cast-border"))
+            .build();
```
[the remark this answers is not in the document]
LordIdra: ItemStackBuilder.gui does that

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/CastingUnit.java
LordIdra: Would it perhaps be better behaviour here to make this a PylonFluidTank and just accept any (castable) fluid? I was very confused at first when nothing happened after I piped liquid copper into the casting unit, and it feels unintuitive from a UX perspective as it's inconsistent with what every other fluid block does

LordIdra (changes requested):
The molding recipes won't work because a clay ball cannot be used with the manual brick mold. May I suggest that instead of the molding recipes, we just do it how tinker's and similar mods do it, and require you to pour liquid gold or similar over a block, ingot, pipe, etc?

Apologies this took me so long to get to

JustAHuman-xD:
Fixes #580

build.gradle.kts
Seggan: yknow what I'ma keep it, it's gonna be api soon anyway

LordIdra (changes requested):
- Casts and casting unit are missing researches
- Pipe cast is actually less efficient than crafting manually - I think we should make it slightly better that crafting manually, maybe like on part with pipe bender since the smeltery is more annoying that pipe benders to automate

Does anyone have other opinions on the cast recipes? The 'craft 9 nugget molds into one nugget mold' thing does not sit right with me at all logically or gameplay-wise - still think it would be better to do something more tinkers-style with casting gold over an ingot or something. is it just me?

Otherwise looks good (pending the switch to using tags for melting points). Have not looked at code since last review but smeltery code is an indecipherable beast to me anyway XD so not going to worry about it as long as everything is good gameplay-wise

Seggan:
The pipe efficiency is exactly the same as crafting them, 6 sheets / 8 outputs = 0.75 ingots per pipe * 144 mb / ingot = 108 mb

I deliberately used the worse ratio of the two I order to not discourage using the pipe bender, as casting pipes is *ridiculously* easy once you have the tech, so I still wanted there to be a reason to use the pipe bender. 

Can you clarify on what you mean "pouring over a gold ingot"? Actually I should probably go look up tinkers lol

Seggan:
To clarify by what I mean "easy": it's literally ingots and diesel in, pipes out, much faster than the pipe bender too

LordIdra:
Here's how it works in tinker's: https://www.youtube.com/watch?v=XjFSR6bF82c

For the pipe casting recipe - mb think I misread the amount. I definitely don't think it's that easy though, like if you're smelting multiple metals with the smeltery you have to keep the right one on the bottom (I saw many people build molten metal filtering setups on the playtest for this), you have to actually fuel the smeltery as you may not have it on constantly (and if you do, that's quite expensive), you need to put the right mold in the caster and then schedule casts. Compared to putting a block of steel into a pipe bender it's quite annoying. I think if it was the same cost as pipe bender it would be much more balanced

LordIdra:
A few remaining issues I found but all small

image
Oh... that's why I thought it was 144mb not 108 lol. This will need updating. (Or removing, given that the amount is defined in the recipe not the mold?)

Pipe and block mold can be placed

Still missing researches for casts (they have the wrong key in researches.yml)

this:
image

LordIdra:
@Seggan just pinging to make sure you saw the above since I realised I didn't submit it as a review ^

Seggan:
Yewh I saw, was just busy with electricity/irl

Seggan:
@LordIdra done

LordIdra:
Pipe/block molds were still placeable but have just fixed myself and merged as it's very small and don't want to block this any longer

PR #864, opened 2026-03-08
Silos

LordIdra:
- Adds silos, single-item storage units
- Adds silo converter, allows you to convert storage units to different types while preserving contents


Seggan (commented):
Doesn't really work in creative but acceptable tradeoff IMO

Seggan (approved):

PR #923, opened 2026-03-27
Change loupe to not break blocks/items in adventure mode

### LINKED ISSUE #908: Disable loupe in adventure mode?
**Issue reported by: cond_b:**
Currently, player can use loupe to break blocks in adventure mode

SchnTgaiSpock:
Closes #908

better solution than straight up disabling it imo

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/content/science/Loupe.java
+++ b/src/main/java/io/github/pylonmc/pylon/content/science/Loupe.java
@@ -263,11 +263,13 @@ public final class Loupe extends RebarItem implements RebarInteractor, RebarCons
 
             new ParticleBuilder(Particle.ITEM).data(stack).extra(0.05).count(16).location(hit.getLocation().add(0, hit.getHeight() / 2, 0)).spawn();
             hit.getWorld().playSound(Sound.sound(SoundEventKeys.ENTITY_ITEM_BREAK, Sound.Source.PLAYER, 0.5f, 1f), hit.getX(), hit.getY(), hit.getZ());
-            if (stack.getAmount() == 1) {
-                hit.remove();
-            } else {
-                stack.subtract();
-                hit.setItemStack(stack);
+            if (player.getGameMode() != GameMode.ADVENTURE) {
+                if (stack.getAmount() == 1) {
+                    hit.remove();
+                } else {
+                    stack.subtract();
+                    hit.setItemStack(stack);
+                }
             }
             addEntry(player, stack.effectiveName(), type.getKey(), getEntryConfig(type));
             player.setCooldown(getStack(), cooldownTicks);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/pylonmc/pylon/content/science/Loupe.java
```diff
@@ -263,11 +263,13 @@ public void onConsumed(@NotNull PlayerItemConsumeEvent event, @NotNull EventPrio
 
             new ParticleBuilder(Particle.ITEM).data(stack).extra(0.05).count(16).location(hit.getLocation().add(0, hit.getHeight() / 2, 0)).spawn();
             hit.getWorld().playSound(Sound.sound(SoundEventKeys.ENTITY_ITEM_BREAK, Sound.Source.PLAYER, 0.5f, 1f), hit.getX(), hit.getY(), hit.getZ());
-            if (stack.getAmount() == 1) {
-                hit.remove();
-            } else {
-                stack.subtract();
-                hit.setItemStack(stack);
+            if (player.getGameMode() != GameMode.ADVENTURE) {
```
OhmV-IR, line 266: This has so many clear bugs. Currently it'll still give the cooldown, play the SFX and particles and add the entry that the player has louped this item even when in adventure mode. I'd recommend adding this as an early return condition in onUsedToClick and sending the player a message when this check fails

OhmV-IR (changes requested):

src/main/java/io/github/pylonmc/pylon/content/science/Loupe.java
SchnTgaiSpock: shouldn't players still be allowed to research in adventure mode? i think it should be up to server admins if they wanna completely disable the loupe for adventure maps

src/main/java/io/github/pylonmc/pylon/content/science/Loupe.java
```diff
@@ -301,7 +303,9 @@ public void onConsumed(@NotNull PlayerItemConsumeEvent event, @NotNull EventPrio
                 }
 
                 hit.getWorld().playEffect(hit.getLocation(), Effect.STEP_SOUND, hit.getBlockData());
-                hit.setType(Material.AIR, true);
+                if (player.getGameMode() != GameMode.ADVENTURE) {
```
OhmV-IR: move this gamemode check up with the type.getHardness check since we don't want to call the BlockBreakEvent when we have no intention of breaking the block

OhmV-IR (changes requested):
alright I guess we can allow louping without necessarily breaking the block / removing the itemstack. Did find something I'd like to fix on my second reread of the code tho.

Seggan (approved):

PR #958, opened 2026-04-18
Fix #957

### LINKED ISSUE #957: Hammer doesn't work with Assembly Table

balugaq:


balugaq:
tested.

Seggan (changes requested):
I can no longer break the table in survival mode

balugaq:
I can use diamond pickaxe to break the table in survival mode, but cannot with hammer.
But it is not supposed to break blocks with hammer, I cannot break any block with hammer actually.

Seggan:
Needs conflicts fixed before I can test, I tried myself but idk hammer code well enough

JustAHuman-xD:
Superseded by #991

PR #971, opened 2026-05-05
[ref:seggan/feature/elektrikity] ELEKTRIKITY

Seggan:
See https://github.com/pylonmc/pylon/issues/987 section "Foundational Pylon PR"


src/main/resources/lang/en.yml
```diff
@@ -2549,6 +2367,172 @@ item:
...
+  boiler:
+    name: "Boiler"
+    lore: |-
+      <arrow> Burns fuel to produce steam
+      <arrow> <attr>Water usage:</attr> %water-usage%
+      <arrow> <attr>Steam production:</attr> %steam-production%
+      
+      <u>Components</u>
+      <diamond> 1x <yellow><item:pylon:boiler/></yellow>
+      <diamond> 20x <yellow><item:pylon:boiler_casing/></yellow>
```
AlmostPanda: 21 Casings

src/main/java/io/github/pylonmc/pylon/util/NumberInputButton.java
```diff
@@ -0,0 +1,91 @@
...
+    private final @Nullable Integer max;
+
+    private final Supplier<Integer> valueGetter;
+    private final Consumer<Integer> valueSetter;
+
+    @lombok.Builder.Default
+    private final Function<Integer, ComponentLike> valueFormatter = Component::text;
+
+    @lombok.Builder.Default
+    private final Consumer<Player> reopenWindow = _unused -> {};
```
JustAHuman-xD: this is unused completely? also no idea why you would need to re-open the window, also is there a reason this is in pylon and not rebar?

src/main/java/io/github/pylonmc/pylon/util/DisplayProjectile.java
```diff
@@ -55,7 +60,10 @@ public DisplayProjectile(
                 .thickness(thickness)
                 .build()
             )
-            .material(material)
+            .itemStack(
+                    ItemStackBuilder.of(material)
+                    .addCustomModelDataString("pylon:display_projectile/" + key)
```
JustAHuman-xD:
```suggestion
                    .addCustomModelDataString("pylon:display_projectile:" + key)
```

src/main/java/io/github/pylonmc/pylon/util/BurnerProgressItem.java
```diff
@@ -0,0 +1,33 @@
...
+import org.bukkit.Material;
+import org.bukkit.entity.Player;
+import org.jetbrains.annotations.NotNull;
+import xyz.xenondevs.invui.item.ItemProvider;
+
+public class BurnerProgressItem extends ProgressItem {
+
+    private static final ItemStackBuilder NOT_BURNING = ItemStackBuilder.of(Material.CHARCOAL)
+            .name(Component.translatable("pylon.gui.burning.not_burning"))
+            .addCustomModelDataString("pylon:burner_progress/not_burning");
```
JustAHuman-xD:
```suggestion
            .addCustomModelDataString("pylon:burner_progress:not_burning");
```

src/main/java/io/github/pylonmc/pylon/util/BurnerProgressItem.java
```diff
@@ -0,0 +1,33 @@
...
+    private static final ItemStackBuilder BURNING = ItemStackBuilder.of(Material.BLAZE_POWDER)
+            .name(Component.translatable("pylon.gui.burning.burning"))
+            .addCustomModelDataString("pylon:burner_progress/burning");
```
JustAHuman-xD:
```suggestion
            .addCustomModelDataString("pylon:burner_progress:burning");
```

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/SmelteryBurner.java
```diff
@@ -109,39 +101,37 @@ public void tick() {
...
     }
 
     @Override
     public void onProcessFinished() {
-        progressItem.setItem(notBurningProgressItem);
-
         Furnace furnace = (Furnace) getBlock().getBlockData();
         furnace.setLit(false);
         getBlock().setBlockData(furnace);
+        refreshBlockTextureItem();
```
JustAHuman-xD:
```suggestion
```

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/SmelteryBurner.java
```diff
@@ -109,39 +101,37 @@ public void tick() {
...
             } else {
-                inventory.setItem(null, i, item.subtract());
+                fuelInventory.setItem(null, i, item.subtract());
             }
 
             startProcess(itemType.getBurnDuration() / 2);
             Furnace furnace = (Furnace) getBlock().getBlockData();
             furnace.setLit(true);
             getBlock().setBlockData(furnace);
+            refreshBlockTextureItem();
```
JustAHuman-xD:
```suggestion
```

src/main/java/io/github/pylonmc/pylon/content/machines/generic/GenericQuarry.java
```diff
@@ -0,0 +1,224 @@
...
+    public void write(@NotNull PersistentDataContainer pdc) {
+        pdc.set(INDEX_KEY, RebarSerializers.INTEGER, index);
+        pdc.set(BLOCK_POSITIONS_KEY, RebarSerializers.LIST.listTypeFrom(RebarSerializers.BLOCK_POSITION), blockPositions);
+        pdc.set(CHUNK_POSITIONS_KEY, RebarSerializers.SET.setTypeFrom(RebarSerializers.CHUNK_POSITION), chunkPositions);
+    }
+
+    private boolean checkBlocks() {
+        while (index < blockPositions.size() - 1) {
+            BlockPosition position = blockPositions.get(index);
+            if (!position.getChunk().isLoaded()) {
```
JustAHuman-xD, line 98: we should totally add a BlockPosition#isLoaded qol method

src/main/java/io/github/pylonmc/pylon/content/machines/generic/GenericPress.java
```diff
@@ -0,0 +1,173 @@
...
+        return Map.of(
+                "input", inputInventory
+        );
+    }
+}
+
+
+
+
+
```
JustAHuman-xD: ?

src/main/java/io/github/pylonmc/pylon/content/machines/generic/GenericMachine.java
```diff
@@ -0,0 +1,169 @@
...
+        if (!outputInventory.canHold(results)) {
+            return true;
+        }
+
+        this.results = results;
+        startRecipe(recipe, getRecipeTicks(recipe));
+        getRecipeProgressItem().setItem(ItemStackBuilder.of(stack.asOne()).clearLore());
+        inputInventory.setItem(new MachineUpdateReason(), 0, stack.subtract(input.getAmount()));
+        return true;
+    }
```
JustAHuman-xD: I would update getResults & how you use it so that you clone the return value of getResults instead of all the implementations supplying clones
1. reduce the tons of .clone() in implementing classes
2. you can clone only when you are sure those are going to be the results which can save memory & tick time

src/main/java/io/github/pylonmc/pylon/util/NumberInputButton.java
```diff
@@ -0,0 +1,91 @@
...
+                RebarArgument.of("increment", valueFormatter.apply(increment)),
+                RebarArgument.of("increment-shift", valueFormatter.apply(shiftIncrement))
+        ));
+        if (min != null) {
+            lore.add(Component.translatable("pylon.gui.number-button.min", RebarArgument.of("min", valueFormatter.apply(min))));
+        }
+        if (max != null) {
+            lore.add(Component.translatable("pylon.gui.number-button.max", RebarArgument.of("max", valueFormatter.apply(max))));
+        }
+        return ItemStackBuilder.gui(material, pylonKey("number_input_button"))
```
JustAHuman-xD: I would say we'd want custom model data specific to each usage of this not a singular "number_input_button"

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/Multimeter.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public void onInteract(@NotNull PlayerInteractEvent event, @NotNull EventPriority priority) {
+        Location playerLocation = event.getPlayer().getLocation();
+        ElectricityPylon pylon = BlockStorage.getByType(ElectricityPylon.class).stream()
+                .min(Comparator.comparing(p -> p.getBlock().getLocation().toCenterLocation().distanceSquared(playerLocation)))
+                .filter(p -> p.getBlock().getLocation().toCenterLocation().distanceSquared(playerLocation) < 64 * 64)
+                .orElse(null);
+        if (pylon == null) return;
+        throw new UnsupportedOperationException("Multimeter interaction not implemented yet");
```
JustAHuman-xD: regardless of how this is changed I just don't think we should every be using getByType like this, instead maybe getting by chunk in a chunk radius loop if you don't change this to just be direct block interaction (which is what I think this should be)

even if you do keep using getByType like this, this will throw an exception for any blocks not in the same dimension as the player, but again I would *not* do this.

src/main/java/io/github/pylonmc/pylon/util/NumberInputButton.java
Seggan: It was used to reopen the window once the player had inputted a specific value in the chat, now that that's been removed this currently sits unused while we switch it over to dialogs

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/SmelteryBurner.java
Seggan: Why?

JustAHuman-xD: the block texture entities now automatically update when block data changes so you don't need to manually refresh it

src/main/java/io/github/pylonmc/pylon/content/machines/generic/GenericQuarry.java
Seggan: This ain't my code lol, I renamed, moved, and modifed `Quarry` and git got confused, thinks I nuked `Quarry`

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/Multimeter.java
Seggan: Oh oops forgot to remove debug code lol

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/SmelteryBurner.java
Seggan: Should probably be documented somewhere

src/main/java/io/github/pylonmc/pylon/PylonPages.java
```diff
@@ -35,6 +35,7 @@ public class PylonPages {
     public static final SimpleStaticGuidePage CARGO = new SimpleStaticGuidePage(pylonKey("machines_cargo"));
     public static final SimpleStaticGuidePage DIESEL_MACHINES = new SimpleStaticGuidePage(pylonKey("machines_diesel_machines"));
     public static final SimpleStaticGuidePage DIESEL_PRODUCTION = new SimpleStaticGuidePage(pylonKey("machines_diesel_production"));
+    public static final SimpleStaticGuidePage ELECTRICITY = new SimpleStaticGuidePage(pylonKey("machines_electricity"));
```
LordIdra: Maybe split this into Electricity Production, Electric Machines, and Electric Utilities?

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/ElectricityPylon.java
```diff
@@ -0,0 +1,47 @@
...
+    public void tick() {
+        ElectricNetwork network = getElectricNodes().getFirst().getNetwork();
+        for (ElectricNode node : network.getNodes()) {
+            Particle.DUST.builder()
+                    .color(Color.fromARGB(network.hashCode()))
+                    .location(node.getBlock().toLocation().toCenterLocation().add(0, 0.6, 0))
+                    .receivers(32, true)
+                    .spawn();
+        }
+    }
```
LordIdra: Forgot to remove?

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/PowerBus.java
LordIdra, line 1: This being a sea lantern feels a bit weird, and the name is also a bit weird considering it's much more a node than a pylon from a player perspective. Maybe worth changing this to use a small cube item display and changing it to 'Electric connector' or similar?

src/main/java/io/github/pylonmc/pylon/PylonItems.java
```diff
@@ -3266,74 +3338,286 @@ private PylonItems() {
...
+        PylonPages.ELECTRICITY.addItem(CAPACITOR_1_KJ);
+    }
+
+    public static final ItemStack WIRE_1_GAUGE = ItemStackBuilder.rebar(Material.STRING, PylonKeys.WIRE_1_GAUGE)
+            .build();
+
+    static {
+        RebarItem.register(PylonWire.class, WIRE_1_GAUGE, PylonKeys.WIRE_1_GAUGE);
+        PylonPages.TOOLS.addItem(WIRE_1_GAUGE);
+    }
```
LordIdra, line 3513: This being in tools doesn't make much sense from a gameplay perspective - I'd suggest moving it to electric machines or components

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/PowerBus.java
LordIdra, line 1: Would it be possible to use six item displays which each span the length of the block, rather than twelve? I know this is quite annoying, but given how many of these are likely to be placed I'm very worried about performance

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/Multimeter.java
LordIdra: Unfinished item?

src/main/resources/lang/en.yml
```diff
@@ -2546,10 +2365,192 @@ item:
...
+    name: "1-Gauge Wire"
+    lore: |-
+      <arrow> Transmits electricity
+      <arrow> <attr>Max power:</attr> %max-power%
+
+  capacitor_1_kj:
+    name: "1 kJ Capacitor"
+    lore: |-
+      <arrow> Stores electricity
+      <arrow> Capacitors have a lower priority than most electric blocks, so they will only take excess power that other blocks can't use
```
LordIdra: The 'lower priority' thing is an implementation detail, not really necessary to include - 'only take excess power' explains it enough

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/generation/Boiler.java
LordIdra, line 1: Missing WAILA

LordIdra, line 1: Would be worth making this not consume fuel when not working, since this was something people complained about with the biorefinery. Or conversely you could go the kiln route and give it a temperature where it has to get over 100 degrees to start working which could be cool and make it more intuitive why it consumes fuel when not working.

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/generation/SteamEngine.java
LordIdra, line 1: Just errors when I place it down...??

src/main/resources/lang/en.yml
```diff
@@ -2546,10 +2365,192 @@ item:
...
+
+  combustion_tower:
+    name: "Combustion Tower"
+    lore: |-
+      <arrow> Burns diesel to produce very hot exhaust
+      <arrow> <attr>Diesel usage:</attr> %diesel-usage%
+      <arrow> <attr>Very hot exhaust production:</attr> %exhaust-production%
+     
+      <u>Components</u>
+      <diamond> 1x <yellow><item:pylon:combustion_tower/></yellow>
```
LordIdra, line 2401: wrong components

src/main/resources/lang/en.yml
```diff
@@ -2546,10 +2365,192 @@ item:
...
+      <u>Components</u>
+      <diamond> 1x <yellow><item:pylon:steam_engine/></yellow>
+      <diamond> 1x <yellow><item:pylon:smokestack_cap/></yellow>
+    waila: "Steam Engine | %bar% | producing %power%"
+
+  gas_turbine:
+    name: "Gas Turbine"
+    lore: |-
+      <arrow> Uses hot gases to produce power
+      <arrow> The more hot gases piped into it, the more power it produces
```
LordIdra, line 2381: Need some more detail here - how much electricity does it produce? what is the efficiency? what is the maximum amount of gas per second?

src/main/resources/lang/en.yml
```diff
@@ -2546,10 +2365,192 @@ item:
...
+      <arrow> <attr>Very hot exhaust production:</attr> %exhaust-production%
+     
+      <u>Components</u>
+      <diamond> 1x <yellow><item:pylon:combustion_tower/></yellow>
+    waila: "Combustion Tower | %diesel% | %exhaust%"
+
+  heat_exchanger:
+    name: "Heat Exchanger"
+    lore: |-
+      <arrow> Exchanges heat between two fluids
```
LordIdra, line 2408: Clueless on what this does from the description

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/generation/GasTurbine.java
LordIdra, line 1: The WAILA just constantly errors when not completed

LordIdra, line 1: The turbine errors and turns into a phantom block on completion???

src/main/resources/lang/en.yml
```diff
@@ -2546,10 +2365,192 @@ item:
...
+      <u>Components</u>
+      <diamond> 1x <yellow><item:pylon:heat_exchanger/></yellow>
+      <diamond> 2x <yellow><item:pylon:fluid_input_hatch/></yellow>
+      <diamond> 2x <yellow><item:pylon:fluid_output_hatch/></yellow>
+
+  wire_1_gauge:
+    name: "1-Gauge Wire"
+    lore: |-
+      <arrow> Transmits electricity
+      <arrow> <attr>Max power:</attr> %max-power%
```
LordIdra, line 2422: Missing instructions

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/machines/ElectricFurnace.java
LordIdra, line 1: Might be worth having the power connection on the top or back here

LordIdra (changes requested):
Stuff I have found from in-game testing so far (have not looked at code much yet):

Almost everything is missing recipes

Almost everything is missing researches

Basically all electric machines are missing WAILAs. (Might be worth rebasing off my PR for this since you'll have to change the WAILAs anyway? I am happy to rebase it and deal with conflicts if you want)

Quarries appear to now be trying to mine air? Not 100% sure if that was introduced by this PR

A lot of the electric machines are iron blocks. I would use some more variety here as it makes them very hard to distinguish, as we do for diesel and hydraulic machines. You could do this in conjunction with some very simple item displays potentially, if you want to keep the base as an iron blocks.

It is also a bit strange that some electric machines have extra item displays (like the press) and some don't

Would be worth adding electric experience bottler (I know this requires a rework of the experience bottler... this is why we should not abstract stuff like we did with the bottler...)

I am not able to test several items in this PR because they immediately cause errors...

src/main/java/io/github/pylonmc/pylon/content/components/FluidHatch.java
LordIdra, line 1: Why aren't we using FluidTankWithDisplayEntity here?

src/main/resources/lang/en.yml
Seggan: Max gas per second is limited only by the hatches, as for electricity that's specified by the gas turbine recipes since it's different per recipe

Seggan:
I mentioned in Discord that I'm splitting the electric content PR into 2: these are just the base, and recipes/researches/more items will come in my next PR

src/main/resources/lang/en.yml
LordIdra: Ah ok. SO presumably it'll be fine once the display recipes are viewable. Might still be worth adding something to hint towards that like 'different gases produce different amounts of power' but idk

src/main/java/io/github/pylonmc/pylon/content/components/FluidHatch.java
Seggan: Cause the original code didn't

LordIdra: The original code worked differently so it didn't need to. The way it works now I think you can take advantage of FluidTankWithDisplayEntity to remove a lot of code

src/main/java/io/github/pylonmc/pylon/util/NumberInputButton.java
LordIdra: +1 to moving this to Rebar

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/generation/Boiler.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+    }
+
+    public static final RebarRegistry<Fuel> FUEL_REGISTRY = new RebarRegistry<>(pylonKey("boiler_fuels"));
+
+    static {
+        for (var fuel : ConfigSection.fromSettings(PylonKeys.BOILER).getOrThrow("fuels", FUELS_TYPE).entrySet()) {
+            FUEL_REGISTRY.register(new Fuel(ItemTypeWrapper.of(fuel.getKey()).getKey(), fuel.getKey(), fuel.getValue()));
+        }
+    }
```
LordIdra: Could we just use vanilla fuels instead of having our own registry, as we do in other blocks now?

Seggan: Yeah this code was made before that change

Seggan: Since youre taking over this machine now Ima mark as resolved and just nuke it

Seggan:
> Quarries appear to now be trying to mine air? Not 100% sure if that was introduced by this PR

This is fine cause `getBreakTicks` returns `null` for air anyway. But I'll exclude it explicity.

Seggan:
Added displays/animations to every iron block machine except brick molder and pipe bender

src/main/resources/lang/en.yml
```diff
@@ -2677,6 +2655,9 @@ guide:
     machines_diesel_machines: "<white>Diesel Machines"
     machines_diesel_production: "<white>Diesel Production"
     machines_smelting: "<white>Smelting"
+    machines_electricity_production: "<yellow>Electricity Production"
+    machines_electric_machines: "<yellow>Electric Machines"
+    machines_electricity_utilities: "<yellow>Electricity Utilties"
```
LordIdra, line 2660: How come these are yellow?

src/main/java/io/github/pylonmc/pylon/recipes/GasTurbineRecipe.java
LordIdra, line 1: It might be good to display power production as its own item, since it's a bit hidden down at the bottom of the turbine lore?

Also:

Image

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/generation/GasTurbine.java
LordIdra, line 1: Having the turbine go from 0 to 60 is a bit jarring, and it stutters if it doesn't have enough input. How much bother would it be to make the speed scale? It doesn't have to be complicated, I'm just thinking constant acceleration/deceleration until stopped / at max speed

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/PowerBus.java
LordIdra, line 1: 'Bus' is the wrong word for this 'hub' maybe?

LordIdra, line 1: What's the use case of this when any number of wires can connect to a node?

I can see a connector being useful but having 6 ports feels strange given the above

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/Capacitor.java
LordIdra, line 1: As I learnt in Quaptics, updating text display text can be really quite expensive, so I'd suggest maybe we limit the text display to only one side and/or make it toggleable? (Since I can imagine if players build a big capacitor bank they might want to turn this off)

LordIdra, line 1: Would be good to show contents in WAILA too

LordIdra (changes requested):
Missing fluid keys

Image

Capacitors do not appear to work

src/main/resources/lang/en.yml
Seggan: ⚡⚡⚡

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/PowerBus.java
Seggan: Wires can't go through blocks/bypassing the wire length limit

LordIdra: Wouldn't it make sense to just have two ports then since they should be able to cover 360 degrees, or am I wrong?

Just looking to reduce entity count as I can imagine these might be used a lot in some bases

Seggan: No youre not wrong

src/main/resources/lang/en.yml
LordIdra: I feel like we shouldn't colour code stuff like that, since well, SF did it and it just ended up as a mess. But it's hardly a big thing for a section ig

src/main/java/io/github/pylonmc/pylon/content/machines/electricity/generation/GasTurbine.java
Seggan: It would get confusing then with visual speed not corresponding to power, and if I scaled power as well that'd be a pretty big rewrite

LordIdra: Something to open an issue about maybe?

It's just really jarring currently

PR #982, opened 2026-05-17
Items can no longer be taken out of a Silo while in creative

### LINKED ISSUE #954: Breaks silo in creative mode also takes an item in it meanwhile.

iTwins:
## The problem
The left click for breaking a silo in creative also took out an item. This was undesirable behavior.

## The solution
Added a creative mode check to the left click for the Silo interact to prevent items from being taken out of a Silo in creative.

Closes #954


LordIdra (approved):
tested

PR #994, opened 2026-05-18
Fix CI

OhmV-IR:
no rm -rfing stuff and just use the intended actions/checkout thing

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -63,13 +63,13 @@ jobs:
 
       - name: Build with Gradle Wrapper
         run: |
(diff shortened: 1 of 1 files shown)


Seggan (commented):
I'm not sure how this would make Gradle use the correct Rebar version without the master project. The only way I can see is if we publish it to local maven and then override Rebar version when building Pylon

OhmV-IR:
> I'm not sure how this would make Gradle use the correct Rebar version without the master project. The only way I can see is if we publish it to local maven and then override Rebar version when building Pylon

?? This still uses the master project, it's not intended to remove it just fix a CI failure that happened when rm -rfing a git cloned directory.

Seggan:
oh whoops misread

Seggan (approved):

PR #996, opened 2026-05-19
Smeltery changes

### LINKED ISSUE #883: Support using lava as fuel in the Smeltery
**Issue reported by: lwj_666:**
why smeltery can't use lava bucket as fuel
why we ignore the hot stuff, the best fuel in vanilla

iTwins:
vanilla fuel for smeltery, removed SmelteryBurnerFuelsPage, fixed smelting multiple items at once, fixed furnaces not lighting.

Closes #883


src/main/java/io/github/pylonmc/pylon/content/machines/smelting/SmelteryBurner.java
```diff
@@ -62,32 +42,23 @@ public SmelteryBurner(@NotNull Block block, @NotNull BlockCreateContext context)
...
-    @Override
-    public void write(@NotNull PersistentDataContainer pdc) {
-        RebarUtils.setNullable(pdc, FUEL_KEY, FUEL_TYPE, fuel);
     }
 
     @Override
     public void postInitialise() {
         setProcessProgressItem(progressItem);
-        createLogisticGroup("fuel", LogisticGroupType.INPUT, inventory);
+        createLogisticGroup("fuel", LogisticGroupType.BOTH, inventory);
```
Seggan, line 55: why both?

iTwins: Because lava buckets leave behind their bucket when consumed. I still wanted it to be automatable even with lava buckets. If you want to keep this logistics group as input only I could make a slot for ejected buckets and make that an output logistics group. That requires a bit more effort, but should be doable.

src/main/java/io/github/pylonmc/pylon/content/machines/smelting/SmelteryBurner.java
```diff
@@ -115,76 +86,52 @@ public void tick() {
...
-                this.fuel = fuel;
-                progressItem.setItem(burningProgressItem);
+            if (item.getType() == Material.LAVA_BUCKET) {
+                inventory.setItem(null, i, new ItemStack(Material.BUCKET));
+            } else {
                 inventory.setItem(null, i, item.subtract());
-                startProcess(Math.round(fuel.burnTimeSeconds * 20));
-                refreshBlockTextureItem();
-                break itemLoop;
             }
```
JustAHuman-xD: this should not be hardcoded, just use ItemType#getCraftingRemainingItem, you should only change it using that if the item is empty afterwards though

iTwins: fixed

LordIdra (approved):

PR #1008, opened 2026-05-26
[ref:human/feature-collection-1] Fix & Feature Collection

### LINKED ISSUE #649: Steel working bug

### LINKED ISSUE #765: Loupe block scan doesn't match rarity
Example: scanning dragon head block will give 1 point compared to 25 by scanning the item

while their name is different: dragon_head_wall vs dragon_head, should we consider fixing this?

Thank you to OMGITSGEORGE ingame for showing us where to find dragon heads

### LINKED ISSUE #802: The Case of the Vanishing Brick Mold
**Issue reported by: overlordidra:**
brick mold recipe *specifically* sometimes just stops displaying in the guide unless viewed indirectly (eg click on diesel brick molder first then brick mold). no errors in console

JustAHuman-xD:
fixes #802 
fixes #649 
fixes #765 

companion to https://github.com/pylonmc/rebar/pull/783
also introduces a SleepingBag implementation but it is currently unused until we handle vanilla multi block blocks

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/content/blocks/SleepingBag.java
+++ b/src/main/java/io/github/pylonmc/pylon/content/blocks/SleepingBag.java
@@ -1,0 +1,77 @@
+package io.github.pylonmc.pylon.content.blocks;
+
+import com.destroystokyo.paper.event.player.PlayerSetSpawnEvent;
+import io.github.pylonmc.pylon.Pylon;
+import io.github.pylonmc.rebar.block.RebarBlock;
+import io.github.pylonmc.rebar.block.base.RebarBed;
+import io.github.pylonmc.rebar.block.context.BlockCreateContext;
+import io.github.pylonmc.rebar.event.PreRebarBlockPlaceEvent;
+import io.github.pylonmc.rebar.event.api.annotation.MultiHandler;
+import org.bukkit.Bukkit;
+import org.bukkit.block.Block;
+import org.bukkit.block.data.type.Bed;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.EventPriority;
+import org.bukkit.event.Listener;
+import org.bukkit.event.player.PlayerBedLeaveEvent;
+import org.bukkit.persistence.PersistentDataContainer;
+import org.jetbrains.annotations.NotNull;
+
+public class SleepingBag extends RebarBlock implements RebarBed {
+    private final Player player;
+
+    public SleepingBag(@NotNull Block block, @NotNull BlockCreateContext context) {
+        super(block, context);
+        this.player = context.getPlayer();
+    }
+
+    public SleepingBag(@NotNull Block block, @NotNull PersistentDataContainer pdc) {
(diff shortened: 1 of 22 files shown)


src/main/java/io/github/pylonmc/pylon/content/blocks/SleepingBag.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public SleepingBag(@NotNull Block block, @NotNull PersistentDataContainer pdc) {
+        super(block, pdc);
+        this.player = null;
+    }
+
+    @Override
+    public void postInitialise() {
+        Bukkit.getScheduler().runTask(Pylon.getInstance(), () -> {
```
LordIdra, line 38: why is this delayed by a tick? (worth commenting?)

src/main/java/io/github/pylonmc/pylon/content/blocks/SleepingBag.java
```diff
@@ -0,0 +1,77 @@
...
+        scheduleBreak();
+    }
+
+    @Override @MultiHandler(priorities = EventPriority.LOWEST, ignoreCancelled = true)
+    public void onSetSpawn(@NotNull PlayerSetSpawnEvent event, @NotNull EventPriority priority) {
+        event.setCancelled(true);
+    }
+
+    private void scheduleBreak() {
+        Bukkit.getScheduler().runTask(Pylon.getInstance(), () -> {
```
LordIdra, line 62: same here

src/main/java/io/github/pylonmc/pylon/content/machines/fluid/FluidTank.java
```diff
@@ -152,6 +154,11 @@ public void onMultiblockUnformed(boolean partUnloaded) {
         }
     }
 
+    @Override
+    public void postBreak(@NotNull BlockBreakContext context) {
+        casings.forEach(FluidTankCasing::reset);
```
LordIdra, line 159: I'm not sure the state should be reset post break, it makes sense that fluid tank casings would stick together even when not with a tank base

Image

Image

LordIdra (changes requested):

src/main/java/io/github/pylonmc/pylon/content/machines/fluid/FluidTank.java
JustAHuman-xD: explained why in vc

LordIdra (approved):

PR #1063, opened 2026-06-12
fix(Rune): Rune effect loss when placed soulbound/fireproof blocks are broken

TagDL:
Fixed a bug where RebarBlocks with Soulbound Rune and Fireproof Rune effects would lose both modifications upon being placed and subsequently broken by players.

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
+++ b/src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
@@ -3,13 +3,14 @@ package io.github.pylonmc.pylon.content.tools;
 import com.destroystokyo.paper.ParticleBuilder;
 import io.github.pylonmc.pylon.content.tools.base.Rune;
 import io.github.pylonmc.rebar.block.RebarBlock;
-import io.github.pylonmc.rebar.block.context.BlockBreakContext.PlayerBreak;
 import io.github.pylonmc.rebar.config.adapter.ConfigAdapter;
 import io.github.pylonmc.rebar.datatypes.RebarSerializers;
 import io.github.pylonmc.rebar.event.RebarBlockBreakEvent;
 import io.github.pylonmc.rebar.event.RebarBlockDeserializeEvent;
 import io.github.pylonmc.rebar.event.RebarBlockPlaceEvent;
 import io.github.pylonmc.rebar.event.RebarBlockSerializeEvent;
+import io.github.pylonmc.rebar.event.RebarBlockUnloadEvent;
+import io.github.pylonmc.rebar.item.RebarItemSchema;
 import io.github.pylonmc.rebar.item.builder.ItemStackBuilder;
 import io.github.pylonmc.rebar.util.RandomizedSound;
 import io.papermc.paper.datacomponent.DataComponentTypes;
@@ -19,7 +20,7 @@ import io.papermc.paper.registry.RegistryKey;
 import io.papermc.paper.registry.keys.tags.DamageTypeTagKeys;
 import io.papermc.paper.registry.tag.Tag;
 import net.kyori.adventure.text.Component;
-import net.kyori.adventure.translation.GlobalTranslator;
+import net.kyori.adventure.text.format.TextDecoration;
 
 import static io.github.pylonmc.pylon.util.PylonUtils.pylonKey;
 
@@ -102,7 +103,7 @@ public class FireproofRune extends Rune {
         ItemStack handle = ItemStackBuilder.of(target.asQuantity(consume)) // Already cloned in `asQuantity`
                 .set(DataComponentTypes.DAMAGE_RESISTANT, DamageResistant.damageResistant(IS_FIRE_TAG))
                 .editPdc(pdc -> pdc.set(FIREPROOF_KEY, RebarSerializers.BOOLEAN, true))
-                .lore(GlobalTranslator.render(TOOLTIP, player.locale()))
+                .lore(TOOLTIP.decoration(TextDecoration.ITALIC, true))
                 .build();
 
         // (N)Either left runes or targets
@@ -139,31 +140,39 @@ public class FireproofRune extends Rune {
     }
 
     public static class FireproofRuneListener implements Listener {
-        private List<RebarBlock> fireproof_blocks = new ArrayList<>();
+        private final List<RebarBlock> fireproofBlocks = new ArrayList<>();
 
         @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
         public void onBlockPlace(RebarBlockPlaceEvent event) {
-            ItemStack itemStack = event.getContext().getItem().asOne();
+            ItemStack itemStack = event.getContext().getItem();
+            if (itemStack == null || itemStack.isEmpty()) return; //for safety
             RebarBlock block = event.getRebarBlock();
             if (itemStack.getPersistentDataContainer().has(FIREPROOF_KEY)) {
-                fireproof_blocks.add(block);
+                fireproofBlocks.add(block);
             }
         }
 
         @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
         public void onBlockBreak(@NotNull RebarBlockBreakEvent event) {
             RebarBlock block = event.getRebarBlock();
-            if (!(event.getContext() instanceof PlayerBreak playerBreak)) return;
-            Player player = playerBreak.event().getPlayer();
-            for (ItemStack itemStack : event.getDrops()) {
-                List<Component> lore = new ArrayList<>(itemStack.lore());
-                if (fireproof_blocks.contains(block)) {
-                    fireproof_blocks.remove(block);
-                    lore.add(GlobalTranslator.render(Component.translatable("pylon.message.fireproof_result.tooltip"), player.locale()));
+            if (fireproofBlocks.contains(block)) {
+                fireproofBlocks.remove(block);
+                for (ItemStack itemStack : event.getDrops()) {
+                    RebarItemSchema dropSchema = RebarItemSchema.fromStack(itemStack);  
+                    if (dropSchema == null) continue; //use schema to make sure as same as itself
+
+                    NamespacedKey blockItemKey = block.getSchema().getKey();
+                    if (blockItemKey == null || !dropSchema.getKey().equals(blockItemKey)) continue;
+
+                    List<Component> lore = new ArrayList<>(itemStack.lore());
+                    lore.add(Component.translatable("pylon.message.fireproof_result.tooltip")
+                        .decoration(TextDecoration.ITALIC, true)); //make sure lore is italic
+                    itemStack.lore(lore);
+
                     itemStack.setData(DataComponentTypes.DAMAGE_RESISTANT, DamageResistant.damageResistant(FireproofRune.IS_FIRE_TAG));
                     itemStack.editPersistentDataContainer(pdc -> pdc.set(FIREPROOF_KEY, RebarSerializers.BOOLEAN, true));
+                    break;
                 }
-                itemStack.lore(lore);
             }
         }
 
@@ -171,7 +180,7 @@ public class FireproofRune extends Rune {
         public void onSerialize(@NotNull RebarBlockSerializeEvent event) {
             RebarBlock block = event.getRebarBlock();
             PersistentDataContainer pdc = event.getPdc();
-            if (fireproof_blocks.contains(block)) {
+            if (fireproofBlocks.contains(block)) {
                 pdc.set(FIREPROOF_KEY, RebarSerializers.BOOLEAN, true);
             }
         }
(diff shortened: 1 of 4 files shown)


JustAHuman-xD (changes requested):
We actually have a system for attaching data directly to blocks, instead of using the chunk pdc like you do here!

If you take a look at some of rebar's built-in interfaces you should be able to find some examples so that you can swap to it. (VirtualInventoryRebarBlock, RecipeProcessorRebarBlock, EntityHolderRebarBlock, to name a few)

Additionally the logic for retaining both runes, soulbound & fireproof should be inside of the corresponding classes in a listener, not inside this abstract class here.

When you move it to both I'd also suggest adding a pdc key to the fireproof rune, and checking for that instead of just the datatag, you could also add some util methods for both which you could use for this. (SoulboundRune#isSoulBound, FireproofRune#isFireproof, for ex)

Overall thank you for the PR! I'm sure with a few changes this can definitely be added

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +137,50 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
                 .count(count)
                 .spawn();
     }
+
+    public static class FireproofRuneListener implements Listener {
+        private List<RebarBlock> fireproof_blocks = new ArrayList<>();
```
JustAHuman-xD: this should be final and called fireproofBlocks

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +137,50 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
...
+
+        @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+        public void onBlockPlace(RebarBlockPlaceEvent event) {
+            ItemStack itemStack = event.getContext().getItem().asOne();
```
JustAHuman-xD: There's no need to do #asOne() here that clones the item stack, you are just checking the PDC so use the raw stack, also iirc getItem is nullable so you may need to check that it isnt null, but I may be wrong

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +137,50 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
...
+                List<Component> lore = new ArrayList<>(itemStack.lore());
+                if (fireproof_blocks.contains(block)) {
+                    fireproof_blocks.remove(block);
+                    lore.add(GlobalTranslator.render(Component.translatable("pylon.message.fireproof_result.tooltip"), player.locale()));
+                    itemStack.setData(DataComponentTypes.DAMAGE_RESISTANT, DamageResistant.damageResistant(FireproofRune.IS_FIRE_TAG));
+                    itemStack.editPersistentDataContainer(pdc -> pdc.set(FIREPROOF_KEY, RebarSerializers.BOOLEAN, true));
+                }
+                itemStack.lore(lore);
+            }
+        }
```
JustAHuman-xD: Honestly I don't know why the runes setting a rendered version of the component, you should be able to just add the Component.translatable results, that also lets you remove the dependency on PlayerBreak and make it apply to any break.

You should also only apply fireproof to the first drop which actually matches the block placed by checking the schemas, not just the first drop.

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +137,50 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
                 .count(count)
                 .spawn();
     }
+
+    public static class FireproofRuneListener implements Listener {
```
JustAHuman-xD: You should listen for RebarBlockUnloadEvent and remove the block from the list when that happens

src/main/java/io/github/pylonmc/pylon/content/tools/SoulboundRune.java
```diff
@@ -78,5 +89,50 @@ public void onPlayerDeath(PlayerDeathEvent event) { // exception being generated
                 }
             }
         }
+
+        private List<RebarBlock> soulbound_blocks = new ArrayList<>();
```
JustAHuman-xD: same as the FireproofRune comments but fields should be at the top of their classes, so this should be above onPlayerDeath

src/main/java/io/github/pylonmc/pylon/content/tools/SoulboundRune.java
JustAHuman-xD: Effectively all my comments for the FireproofRune version of this apply to SoulboundRune as well

JustAHuman-xD (changes requested):

LordIdra:
Tested that this solution fixes rebar blocks, but it does not address vanilla blocks and we definitely should as it's more or less the same issue.

Unfortunately I think we would need to switch back to using chunk's PDC to fix this for normal blocks, not using the approach Justin suggested.

@JustAHuman-xD do you have any input since you reviewed it originally?

LordIdra:
> Tested that this solution fixes rebar blocks, but it does not address vanilla blocks and we definitely should as it's more or less the same issue.
> 
> Unfortunately I think we would need to switch back to using chunk's PDC to fix this for normal blocks, not using the approach Justin suggested.
> 
> @JustAHuman-xD do you have any input since you reviewed it originally?

Discussed in VC - I think using the chunk's PDC is still the best way forward. Could you change it back to that and handle the vanilla case? Sorry for the confusion

TagDL:
> > Tested that this solution fixes rebar blocks, but it does not address vanilla blocks and we definitely should as it's more or less the same issue.
> 
> > 
> 
> > Unfortunately I think we would need to switch back to using chunk's PDC to fix this for normal blocks, not using the approach Justin suggested.
> 
> > 
> 
> > @JustAHuman-xD do you have any input since you reviewed it originally?
> 
> 
> 
> Discussed in VC - I think using the chunk's PDC is still the best way forward. Could you change it back to that and handle the vanilla case? Sorry for the confusion

Already change it to chunk's PDC❤️

src/main/java/io/github/pylonmc/pylon/content/tools/SoulboundRune.java
```diff
@@ -65,18 +71,120 @@ public void onContactItem(@NotNull PlayerDropItemEvent event, @NotNull ItemStack
...
+     * @return true if the soulbound rune has been used on the item, false otherwise
+     */
+    public static boolean hasRuneApplied(@NotNull ItemStack itemStack) {
+        return itemStack.getPersistentDataContainer().getOrDefault(SOULBOUND_KEY, RebarSerializers.BOOLEAN, false);
+    }
+
     public static class SoulboundRuneListener implements Listener {
+
         @EventHandler
         public void onPlayerDeath(PlayerDeathEvent event) { // exception being generated
```
balugaq:
```suggestion
        @EventHandler(ignoreCancelled = true)
        public void onPlayerDeath(PlayerDeathEvent event) { // exception being generated
```

src/main/java/io/github/pylonmc/pylon/content/tools/SoulboundRune.java
```diff
@@ -65,18 +71,120 @@ public void onContactItem(@NotNull PlayerDropItemEvent event, @NotNull ItemStack
...
+            }
+            ItemStack pdcItemStack = block.getChunk().getPersistentDataContainer()
+                    .get(getChunkPDC(block.getLocation()), RebarSerializers.ITEM_STACK);
+
+            if (block.getState() instanceof ShulkerBox box) {
+                event.setDropItems(false);
+                BlockStateMeta meta = (BlockStateMeta) pdcItemStack.getItemMeta();
+
+                meta.setBlockState(box);
+                pdcItemStack.setItemMeta(meta);
```
balugaq: I would prefer use https://jd.papermc.io/paper/26.2/io/papermc/paper/datacomponent/DataComponentTypes.html#CONTAINER

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +123,107 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
...
+            }
+
+            ItemStack pdcItemStack = block.getChunk().getPersistentDataContainer()
+                    .get(getChunkPDC(block.getLocation()), RebarSerializers.ITEM_STACK);
+            for (Item item : event.getItems()) {
+                ItemStack itemStack = item.getItemStack();
+                if (pdcItemStack.getType() != itemStack.getType()) {
+                    continue;
+                }
+                event.getItems().remove(item);
```
balugaq:
```suggestion
                itemStack.subtract();
```

balugaq: Fix breaking nether warts only drops 1. ( planted fireproofed `NETHER_WARTS` (which has the same `Materal` whether to be block or to be item.)

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +123,107 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
...
+
+            if (!block.getChunk().getPersistentDataContainer().has(getChunkPDC(block.getLocation()))) {
+                return;
+            }
+
+            ItemStack pdcItemStack = block.getChunk().getPersistentDataContainer()
+                    .get(getChunkPDC(block.getLocation()), RebarSerializers.ITEM_STACK);
+            for (Item item : event.getItems()) {
+                ItemStack itemStack = item.getItemStack();
+                if (pdcItemStack.getType() != itemStack.getType()) {
```
balugaq: You can't just use material type to check them. Otherwise there's a dupe:

https://github.com/user-attachments/assets/4ab1ef35-4841-4eb2-8e07-fd0ab13c7011

src/main/java/io/github/pylonmc/pylon/content/tools/SoulboundRune.java
```diff
@@ -65,18 +71,120 @@ public void onContactItem(@NotNull PlayerDropItemEvent event, @NotNull ItemStack
...
+            }
+
+            ItemStack pdcItemStack = block.getChunk().getPersistentDataContainer()
+                    .get(getChunkPDC(block.getLocation()), RebarSerializers.ITEM_STACK);
+            for (Item item : event.getItems()) {
+                ItemStack itemStack = item.getItemStack();
+                if (pdcItemStack.getType() != itemStack.getType()) {
+                    continue;
+                }
+                event.getItems().remove(item);
```
balugaq: same

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +123,107 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
...
+            }
+            ItemStack pdcItemStack = block.getChunk().getPersistentDataContainer()
+                    .get(getChunkPDC(block.getLocation()), RebarSerializers.ITEM_STACK);
+
+            if (block.getState() instanceof ShulkerBox box) {
+                event.setDropItems(false);
+                BlockStateMeta meta = (BlockStateMeta) pdcItemStack.getItemMeta();
+
+                meta.setBlockState(box);
+                pdcItemStack.setItemMeta(meta);
```
balugaq: same

balugaq (changes requested):
There's a lot of duplicate code in FireproofRune.Listener and SouldboundRune.Listener. I would prefer to add a manager to handle them.

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
balugaq: Same for the rebar blocks.

src/main/java/io/github/pylonmc/pylon/content/tools/FireproofRune.java
```diff
@@ -106,4 +108,38 @@ public void spawnParticle(@NotNull Particle particle, @NotNull Location location
...
+        @Override
+        protected boolean hasRuneApplied(@NotNull ItemStack stack) {
+            return FireproofRune.hasRuneApplied(stack);
+        }
+
+        @Override
+        protected @NotNull ItemStack applyRune(@NotNull ItemStack stack, int amount) {
+            return FireproofRune.applyRune(stack, amount);
+        }
+    }
```
LordIdra, line 144: I think that instead of having a class per rune which extends the listener, you can just have one listener in the `Rune` class and make `hasRuneApplied` and `applyRune` and `getKeyPrefix` abstract methods in `Rune` itself. (But let me know if I am overlooking something, or you need more details)

src/main/java/io/github/pylonmc/pylon/content/tools/base/Rune.java
```diff
@@ -114,4 +127,112 @@ void onRuneDrop(@NotNull PlayerDropItemEvent event) {
...
+            if (!block.getChunk().getPersistentDataContainer().has(pdc)) {
+                return;
+            }
+
+            ItemStack pdcItemStack = block.getChunk().getPersistentDataContainer()
+                    .get(pdc, RebarSerializers.ITEM_STACK);
+
+            if (pdcItemStack.hasData(DataComponentTypes.CONTAINER)) { //prevent the appearance of dupe
+                pdcItemStack.setData(DataComponentTypes.CONTAINER, ItemContainerContents.containerContents(new ArrayList<>()));
+            }
```
LordIdra, line 189: Could you improve the comment here? I don't understand what this code is for from the existing code + comment

LordIdra (changes requested):
Apologies it has taken a while to get back to this

PR #1069, opened 2026-06-19
Change event priority for Lumber Axe block break handler

TagDL:
fix lumberAxe will drop rebar block and log at the same time


JustAHuman-xD:
Superceded by #1071

PR #1084, opened 2026-07-18
fix core drill throwing error

### LINKED ISSUE #1010: Core drills throw errors every tick if they don't have a block below the drill bit
Image

OhmV-IR:
Fixes #1010 
The issue is that Material.AIR.isItem() returns true because there is an Air ItemType but this is not a valid type for spawning particles. Fixed by adding an additional check to the block to make sure that the material is not air and has an item type


balugaq (approved):

Seggan (approved):

LordIdra (approved):
tested

PR #1085, opened 2026-07-20
add extra effect to mixing pot

### LINKED ISSUE #1039: extra effects of mix pot
**Issue reported by: lwj_666:**
[View original discord thread](https://discord.com/channels/1329177304857055273/1511320693940424765)
when player fall into a mix pot with:
- cold fluid: freeze (like powder snow crucible)
- hot fluid: on fire (like lava crucible)

TagDL:
Resolves #1039

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
+++ b/src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
@@ -10,23 +10,33 @@ import io.github.pylonmc.rebar.block.RebarBlock;
 import io.github.pylonmc.rebar.block.interfaces.CauldronRebarBlockHandler;
 import io.github.pylonmc.rebar.block.interfaces.DirectionalRebarBlock;
 import io.github.pylonmc.rebar.block.interfaces.InteractRebarBlockHandler;
+import io.github.pylonmc.rebar.block.interfaces.TickingRebarBlock;
 import io.github.pylonmc.rebar.block.context.BlockCreateContext;
 import io.github.pylonmc.rebar.event.api.annotation.MultiHandler;
 import io.github.pylonmc.rebar.fluid.FluidPointType;
 import io.github.pylonmc.rebar.fluid.RebarFluid;
+import io.github.pylonmc.rebar.fluid.tags.FluidTemperature;
 import io.github.pylonmc.rebar.i18n.RebarArgument;
 import io.github.pylonmc.rebar.item.RebarItem;
 import io.github.pylonmc.rebar.recipe.ingredient.FluidOrItem;
 import io.github.pylonmc.rebar.recipe.ingredient.FluidWithAmount;
 import io.github.pylonmc.rebar.recipe.ingredient.ItemChoice;
 import io.github.pylonmc.rebar.util.ProgressBar;
+import io.github.pylonmc.rebar.util.RandomizedSound;
 import io.github.pylonmc.rebar.util.gui.unit.UnitFormat;
 import io.github.pylonmc.rebar.waila.WailaDisplay;
+import io.papermc.paper.registry.keys.SoundEventKeys;
+import kotlin.Pair;
+import net.kyori.adventure.sound.Sound;
+
 import org.bukkit.Material;
 import org.bukkit.Particle;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.damage.DamageSource;
+import org.bukkit.damage.DamageType;
 import org.bukkit.entity.Item;
+import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Event;
 import org.bukkit.event.EventPriority;
@@ -44,6 +54,7 @@ import java.util.List;
 
 public final class MixingPot extends RebarBlock implements
         DirectionalRebarBlock,
+        TickingRebarBlock,
         InteractRebarBlockHandler,
         FluidTankWithDisplayEntity,
         CauldronRebarBlockHandler {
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -126,6 +138,37 @@ public void onInteractedWith(@NotNull PlayerInteractEvent event, @NotNull EventP
         }
     }
 
+    @Override
+    public void tick() {
+        if (getFluidType() == null) {
+            if (getBlock().getType() != Material.CAULDRON) {
+                getBlock().setType(Material.CAULDRON);
+            }
```
balugaq: Block type never changes.
```suggestion
```

TagDL: Damn i forgot to delete it

src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -126,6 +138,33 @@ public void onInteractedWith(@NotNull PlayerInteractEvent event, @NotNull EventP
         }
     }
 
+    @Override
+    public void tick() {
+
+            return;
+        }
```
balugaq:
```suggestion
```
oh no, just fix compile

Seggan (commented):
Why is freeze ticks dependent on tick interval but not fire ticks?

TagDL:
> Why is freeze ticks dependent on tick interval but not fire ticks?

Freeze ticks represent the degree of freezing. Making the freezing effect dependent on the tick interval is intended to simulate the player gradually becoming frozen after entering powder snow. But when a player enters lava, they immediately catch fire, and their fire duration is reset to 300 ticks.

src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -67,6 +78,7 @@ public MixingPot(@NotNull Block block, @NotNull BlockCreateContext context) {
         super(block, context);
         createFluidDisplay();
         setFacing(context.getFacing());
+        setTickInterval(1);
```
Seggan: Needs to be configurable (see literally every other machine)

Seggan (approved):

Seggan (changes requested):
Woops wrong button

Seggan (approved):

src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -126,6 +141,30 @@ public void onInteractedWith(@NotNull PlayerInteractEvent event, @NotNull EventP
         }
     }
 
+    @Override
+    public void tick() {
```
LordIdra, line 145: Before ticking, you should check if the mixing pot holds a HOT or COLD fluid to avoid unnecessary ticking

src/main/resources/settings/mixing_pot.yml
```diff
@@ -0,0 +1 @@
+tick-interval: 1
```
LordIdra, line 1: This means that the mixing pot will tick 20 times per second, which is very fast (most other machines have intervals of 10-40). This is a performance concern as `getNearbyLivingEntities` can be a bit laggy and the mixing pot is a quite common block. I would suggest increasing this to 5 or 10 ticks

src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -126,6 +141,30 @@ public void onInteractedWith(@NotNull PlayerInteractEvent event, @NotNull EventP
...
+        for (LivingEntity entity : getBlock().getLocation().toCenterLocation().getNearbyLivingEntities(0.4)) {
+            if (getFluidType().getTag(FluidTemperature.class) == FluidTemperature.HOT) {
+                entity.setFireTicks(300);
+                entity.damage(4.0F, DamageSource.builder(DamageType.LAVA).build());
```
LordIdra, line 149: The damage amount should probably be configurable, and should take into account the tick rate. For example if the tick interval is 1, then this would damage 4*20 = 80 hearts per second which is probably not what we want

src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -126,6 +141,30 @@ public void onInteractedWith(@NotNull PlayerInteractEvent event, @NotNull EventP
...
+
+                new RandomizedSound(List.of(SoundEventKeys.ENTITY_GENERIC_BURN), //play sound like vanilla cauldron
+                        Sound.Source.NEUTRAL,
+                        new Pair<>(0.4, 0.4),
+                        new Pair<>(2.0, 2.4))
+                    .playAt(entity);
```
LordIdra, line 155: This can also be made configurable as we have a sound adaptor - see the docs for that to see how to format the config

src/main/java/io/github/pylonmc/pylon/content/machines/simple/MixingPot.java
```diff
@@ -126,6 +141,30 @@ public void onInteractedWith(@NotNull PlayerInteractEvent event, @NotNull EventP
...
+
+            } else if (getFluidType().getTag(FluidTemperature.class) == FluidTemperature.COLD) {
+                int freezeTick = Math.min(entity.getFreezeTicks() + getTickInterval() + 2, entity.getMaxFreezeTicks());
```
LordIdra, line 158: Where does the `+ 2` come from?

LordIdra (changes requested):

PR #1089, opened 2026-07-24
Fix xp dupe

### LINKED ISSUE #1070: xp dupe
**Issue reported by: thefuturepickle:**
[View original discord thread](https://discord.com/channels/1329177304857055273/1519758196066095165)
xp talisman affects the xp coming out of the xp sprinkler

balugaq:
Fixes #1070


Seggan (approved):

LordIdra (approved):
I can see this potentially causing issues where the talisman would not multiply XP which comes from e.g. commpleting a quest for a quest plugin. But there is no way really for us to tell whether the XP should be allowed to multiply or not so I think this is the best solution.

PR #1093, opened 2026-08-18
Fix missing sleeping bag translation

wickidcow:
Adds the missing English translation for the Sleeping Bag.

I noticed the item did not have an entry in en.yml, causing the translation to be missing when using the item in-game.

This only adds the missing Sleeping Bag name and does not change any item behavior.

### DIFF
--- a/src/main/resources/lang/en.yml
+++ b/src/main/resources/lang/en.yml
@@ -1,6 +1,9 @@
 addon: "Pylon"
 
 item:
+  sleeping_bag:
+    name: "Sleeping Bag"
+
   bandage:
     name: "Bandage"
     lore: |-


Seggan:
Thank you for the PR. The sleeping bag translation is intentionally missing as we have not finished the implementation yet (IIRC we were waiting on a Paper PR). More relevantly, Minecraft 26.3 adds the straw bed which basically is the same thing, so we are going to remove the sleeping bag completely.

wickidcow:
Oh! Ok. I'd presume the hay bed would only work one time? But maybe it would be good to have it still if that's the case

Seggan:
idk how the straw bed works tbh but IMO it's still not worth having the sleeping bad, ik @JustAHuman-xD agrees with me

PR #1094, opened 2026-08-18
Register with Rebar before initializing metrics

wickidcow:
Moves Pylon's Rebar registration before bStats initialization.

I ran into this while testing startup failures between Pylon and Rebar. If Rebar rejects the addon during startup, Pylon should not initialize additional plugin services first. 

This keeps Pylon from being partially initialized when registerWithRebar() fails.


Seggan (approved):
Unnecessary comment but ok

PR #1096, opened 2026-08-18
Fix Rebar ref lookup in CI

### LINKED ISSUE #1095: CI ignores the Rebar ref specified in PR titles
### Description

The Gradle workflow defines the Rebar ref parser step with the ID `parseref`, but the Rebar checkout step reads `steps.rebarcommit.outputs.ref`:

```yaml
- name: Get commit of rebar to use
  id: parseref
  # ...

- name: Clone rebar
  uses: actions/checkout@v5
  with:
    repository: 'pylonmc/rebar'
    ref: ${{ steps.rebarcommit.outputs.ref }}
```

There is no step with the ID `rebarcommit`, so the expression resolves to an empty value and checkout falls back to Rebar's default branch. As a result, a pull request whose title starts with `[ref:<branch>]` is not tested against the requested Rebar branch.

### Expected behavior

The checkout step should use the output of the parser step:

```yaml
ref: ${{ steps.parseref.outputs.ref }}
```

This matters for changes developed across Pylon and Rebar: CI can otherwise report misleading failures or successes because it builds against the wrong Rebar revision.

ddism:
## Summary

- read the requested Rebar branch from the existing `parseref` step
- restore the `[ref:<branch>]` PR title behavior used for cross-repository changes

## Testing

- `git diff --check`
- verified that the workflow output reference matches the declared step ID

Fixes #1095

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -46,7 +46,7 @@ jobs:
         uses: actions/checkout@v5
         with:
           repository: 'pylonmc/rebar'
-          ref: ${{steps.rebarcommit.outputs.ref}} # empty string if ref not found falls back to master
+          ref: ${{steps.parseref.outputs.ref}} # empty string if ref not found falls back to master
           path: 'parallel-dev-repo/rebar'
 
       - name: Clone pylon


Seggan (approved):

LordIdra:
Thanks for the pr :+1:

PR #1102, opened 2026-09-02
Clean up Paper 26.2 deprecations

### LINKED ISSUE #1099: 26.2 deprecations
```
/home/seggan/Progs/Kotlin/pylonmc/pylon/src/main/java/io/github/pylonmc/pylon/content/science/Loupe.java:313: warning: [removal] STEP_SOUND in Effect has been deprecated and marked for removal
                hit.getWorld().playEffect(hit.getLocation(), Effect.STEP_SOUND, hit.getBlockData());
                                                                   ^
/home/seggan/Progs/Kotlin/pylonmc/pylon/src/main/java/io/github/pylonmc/pylon/content/tools/LumberAxe.java:66: warning: [removal] STEP_SOUND in Effect has been deprecated and marked for removal
        block.getWorld().playEffect(block.getLocation(), Effect.STEP_SOUND, block.getBlockData());
                                                               ^
/home/seggan/Progs/Kotlin/pylonmc/pylon/src/main/java/io/github/pylonmc/pylon/util/DisplayProjectile.java:115: warning: [removal] EntityDamageEvent(@org.jetbrains.annotations.NotNull Entity,@org.jetbrains.annotations.NotNull DamageCause,@org.jetbrains.annotations.NotNull DamageSource,double) in EntityDamageEvent has been deprecated and marked for removal
            EntityDamageEvent event = new EntityDamageEvent(
                                      ^
```

wickidcow:
Cleans up the remaining Paper 26.2 deprecation warnings from #1099.

* Replaces `Effect.STEP_SOUND` with `Effect.DESTROY_BLOCK` for the loupe break effect.
* Uses `Damageable#damage` instead of constructing `EntityDamageEvent` directly for display projectiles.
* Keeps knockback and hit sounds from firing when damage is cancelled.

And lastly the `LumberAxe` warning listed in the issue is already gone on current master.

Closes #1099

### DIFF
--- a/src/main/java/io/github/pylonmc/pylon/util/DisplayProjectile.java
+++ b/src/main/java/io/github/pylonmc/pylon/util/DisplayProjectile.java
@@ -14,7 +14,6 @@ import org.bukkit.entity.Damageable;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.ItemDisplay;
 import org.bukkit.entity.Player;
-import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.util.Vector;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/pylonmc/pylon/util/DisplayProjectile.java
LordIdra, line 1: The issue here is that the constructor we were using has been deprecated - not the event itself. The event should still be fired as we need other plugins to know that we are damaging the entity, and calling `damage` does not fire the event automatically AFAIK. You should change the code to work the same way as it did before but using a different constructor.

LordIdra (changes requested):
Thanks for the PR

Tested build no longer gives the relevant deprecation warnings

PR #1105, opened 2026-09-06
fix: update crude alloy furnace lit state

### LINKED ISSUE #1009: crude alloy furnace missing lit model data
See kiln for how to do this

chrisriv10:
## Issue

Fixes #1009

The Crude Alloy Furnace appeared unlit regardless of whether or not it was burning fuel.

## Summary

Updated the Crude Alloy Furnace so its block state correctly shows whether it currently has fuel. This makes the furnace appear lit while burning fuel and unlit when fuel runs out. 

## Testing

- Ran `./gradlew compileJava` successfully against the current Rebar.
- Tested the Crude Alloy Furnace in game.
- Before the change, the furnace processed recipes but never lit.
- After the change, the furnace now lights up while fuel is burning and turns off when fuel runs out.


Seggan (approved):

LordIdra (approved):
thanks for the PR

